PR #8, opened 2017-04-30
Add `sheet.getStyleElement()` to the demo

jameslnewell:
I just upgraded the SSR API to the latest RC. I couldn't figure out how to get the generated CSS into the `<head>` when using `ReactDOM.renderToStaticMarkup()` and was about to try PR'ing this exact same method. Would be good to document it somewhere 😁


philpl:
It's actually mentioned in the new text, but I'll merge it anyway, since this makes it twice as clear :)

PR #11, opened 2017-05-12
reduce repaints on fixed sidebar

thysultan:
`position: fixed` causes the sidebar to get repainted on the CPU with every scroll event, small hack to delegate this to the GPU.


philpl:
Cheers :)

Issue #10, opened 2017-05-11
Unable to scroll the documentation menu when at a small breakpoint 


superhighfives:
When visiting https://sc-next-docs.philpl.com/docs at a smaller breakpoint, the overlay menu doesn't allow scrolling on desktop (OSX, Chrome 57).

On mobile (iOS 10, Safari) it seems a bit touch and go whether you're scrolling the overlay menu or the body itself.

![screen shot 2017-05-11 at 14 05 16](https://cloud.githubusercontent.com/assets/449385/25951589/143acb98-3656-11e7-8bd9-7330fb88b8e8.png)

philpl:
Thanks for reporting! I'll take a look

PR #30, opened 2017-05-24
#10 Resolved the scroll issue when changing breakpoints

superhighfives:
Hey @philpl—I've had a look at the issue, and you're right, it does appear to be the way you're managing scrolling.

I've added a couple of fixes, but keen to get your feedback on whether there's a more "styled-components" way of doing it. Especially when attaching classes to the body.

- Added a sticky class to the body when the navigation is expanded, which adds `overflow: hidden;`
- Added event listeners when loading on the desktop breakpoint, to trigger when resizing between desktop and mobile. This shouldn't impact mobile.

Happy to discuss.

### DIFF
--- a/components/CaptureScroll.js
+++ b/components/CaptureScroll.js
@@ -61,13 +61,24 @@ const captureScroll = Component => {
       }
     }
 
+    onResize = evt => {
+      isMobile = window.matchMedia(`(max-width: ${1000 / 16}em)`).matches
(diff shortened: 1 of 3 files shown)


philpl:
It seems fine for now :) Let's ship this and see if there's a better solution in the future ;)

superhighfives:
Awesome!

PR #31, opened 2017-05-24
Slightly fade out the logos on the homepage

mxstbr:
Before they would be the first thing that would hit your eye, which is
suboptimal as we've got a great tagline AND a live editor

Fixed that by making them `opacity: 0.6`, but I'm open to other
solutions. Making them white would be kinda awesome maybe? Any clues how
to do that?


mxstbr:
I made 'em white! :tada:

screen shot 2017-05-25 at 11 09 10

PR #34, opened 2017-05-25
drop pixels from input size

rajington:
Input sizes are [numbers](https://www.w3schools.com/tags/att_input_size.asp), I'm not sure if that was the intention but changing a pixel value makes it a little confusing.


geelen:
Wow, yeah, I haven't used `size` in forever, but you're totally right.

PR #35, opened 2017-05-26
Migration informations for RN users

MatthieuLemoine:
This PR updates RN examples & add migration informations for v1 users.

See https://github.com/styled-components/styled-components/issues/835 for context.

### DIFF
--- a/components/basics/react-native.js
+++ b/components/basics/react-native.js
@@ -1,7 +1,9 @@
 import React from 'react'
 import SectionLayout from '../SectionLayout'
 import CodeBlock from '../CodeBlock'
-import Code from '../Code'
+import Code from '../Code';
+import Note from '../Note';
+import Link from '../Link';
 
 const sample = (`
 import styled from 'styled-components/native';
@@ -28,9 +30,9 @@ class MyReactNativeComponent extends React.Component {
 const cssToReactNativeSample = (`
 const RotatedBox = styled.View\`
   transform: rotate(90deg);
-  text-shadow-offset: 10 5;
+  text-shadow-offset: 10px 5px;
   font-variant: small-caps;
-  margin: 5 7 2;
+  margin: 5px 7px 2px;
 \`;
 `).trim()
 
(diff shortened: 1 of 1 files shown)


philpl:
Let me take a look at how this looks "in person" :)

MatthieuLemoine:
Not sure about how links should be styled

philpl:
I'd prefer to remove the example and just turn this into a `<Note>`. The links are supposed to use the `Link` component with the `inline` prop ;) If you'd like, I can make the appropriate changes later

components/basics/react-native.js
```diff
@@ -28,9 +28,29 @@ class MyReactNativeComponent extends React.Component {
...
   transform: rotate(90deg);
-  text-shadow-offset: 10 5;
+  text-shadow-offset: 10px 5px;
   font-variant: small-caps;
-  margin: 5 7 2;
+  margin: 5px 7px 2px;
+\`;
+`).trim()
+
+const migrationSample = (`
```
philpl: Don't need an extensive example if we can write a precise note and link to the codemod :)

components/basics/react-native.js
```diff
@@ -60,6 +80,18 @@ const ReactNative = () => (
       the <Code>keyframes</Code> and <Code>injectGlobal</Code> helpers since React Native doesn't
       support keyframes or global styles. We will also warn you if you use media queries or nest your CSS.
     </p>
+
+    <p>
+      Since v2, <Code>css-to-react-native</Code> supports percentages. To make this possible we had to enforce unit usage when the value is a length.
```
philpl: This should just be a `<Note>`. Also, a bit of a nit on wording: `In v2`, since there's been a lot of versions of it.

- Don't need to mention the sub-library `css-to-react-native`
- Don't need to split this up and make it longer

e.g. "In v2 we support percentages. To make this possible we need to enforce units for all shorthands. If you're migrating to v2, [a codemod is available]." <- [] for link

components/basics/react-native.js
```diff
@@ -60,6 +80,18 @@ const ReactNative = () => (
...
+      See <Code><a href="https://github.com/styled-components/css-to-react-native/issues/20">css-to-react-native#20</a></Code> for more informations.
```
philpl: Don't need a link to the issue, as the explanation fits in there and can be linked to from the codemod

components/basics/react-native.js
```diff
@@ -60,6 +80,18 @@ const ReactNative = () => (
...
+    </p>
+
+    <CodeBlock code={migrationSample} language="jsx" />
+
+    <p>
+      For the v1 users, a <Code><a href="https://github.com/styled-components/styled-components-native-code-mod">codemod</a></Code> is available to do this conversion for you.
```
philpl: Can be wrapped into the note (see above)

MatthieuLemoine:
You're right it looks nicer now.

PR #38, opened 2017-05-28
Fixed StyleSheetManager doesn't get the sheet.instance

sergop321:
https://github.com/styled-components/styled-components/issues/453#issuecomment-298502900 Updated the docs


mxstbr:
I think this should be fixed in the library? /cc @philpl @geelen

philpl:
@mxstbr I think we should aim for a general refactor of the implementation of the stylesheets, since I think we can make everything more explicit and smaller :) it is probably unfavourable to copy global rules in this way as well, and instead go for referencing an instance of stylesheet if it's possible.

For now we should definitely just update the docs :)

PR #41, opened 2017-06-02
Feature/#12 operator mono for code

superhighfives:
Right! Updated the fonts because cloud.typography is being difficult.


superhighfives:
Thank @philpl! Can you let me know when you've deployed?

PR #42, opened 2017-06-02
Feature/faq

oliverturner:
Inspired by a conversation [here](https://twitter.com/timarney/status/870269647656022017) this adds a couple of sections under a FAQ/Tips and Tricks heading.


timarney:
Another question for this maybe.
What's best practice for property reuse I.e Sass mixins or extending /sharing styles between components.

pages/docs.json
```diff
@@ -62,6 +62,13 @@
           "title": "TypeScript"
         }
       ]
+    }, {
+      "title": "FAQ's and Tips & Tricks",
```
philpl: Can you rename it to be shorter maybe :/ We don't really need to call it "Tips & Tricks". We can just reword all the subsections to be questions.

e.g. "How do I nest selectors?"

localhost-3000-docs 2

pages/docs.json
```diff
@@ -62,6 +62,13 @@
...
+      "pathname": "faqs",
```
philpl: Can we rename this to just be `faq` please? :) Also above, we can just use "FAQs" (no apostrophe)

components/faqs/nesting.js
```diff
@@ -0,0 +1,44 @@
...
+render(
+  <ColorChanger href="#">
+    <h2>Hello world!</h2>
+  </ColorChanger>
+)
+`).trim()
+
+const Nesting = () => (
+  <SectionLayout title="Nesting">
+    <p>Nesting is a feature ported from Sass. Used sparingly it's a great way to lighten your code by reducing the need to create explicit classes for every element.</p>
```
philpl: This only uses a direct descendent selector and doesn't really talk about how an interpolation can be used as well :)

components/faqs/reverse-selectors.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+\`;
+
+render(
+  <Link href="#">
+    <Icon viewBox="0 0 512 512">
+      <path d="M233.288,197.742l-49.948-48.33c-3.136-3.034-7.999-3.377-11.53-0.811l-10.189,7.409l-36.874-50.718l27.151-19.74 c2.032-1.477,3.353-3.737,3.642-6.233c0.289-2.495-0.479-4.997-2.12-6.9L100.1,10.565C94.312,3.851,85.904,0,77.032,0    c-6.466,0-12.653,2.012-17.89,5.82L35.818,22.779c-12.708,9.239-16.272,26.995-8.116,40.425l42.397,69.793    c1.305,2.147,3.448,3.649,5.91,4.143c0.587,0.119,1.179,0.175,1.769,0.175c1.883,0,3.736-0.592,5.283-1.718l27.15-19.739    l36.875,50.717l-10.189,7.409c-3.53,2.567-4.705,7.299-2.785,11.218l30.571,62.418c1.168,2.384,3.328,4.129,5.904,4.768    c0.715,0.177,1.441,0.265,2.165,0.265c1.88,0,3.733-0.591,5.284-1.718l54.289-39.47c2.147-1.561,3.493-3.99,3.679-6.637    C236.19,202.18,235.195,199.587,233.288,197.742z M106.918,96.038c-0.002,0.001-0.005,0.002-0.007,0.005    c-0.002,0.001-0.004,0.004-0.006,0.005l-26.516,19.279l-37.33-61.451c-3.342-5.502-1.882-12.778,3.325-16.563l23.324-16.958 c2.145-1.561,4.679-2.385,7.324-2.385c3.638,0,7.086,1.578,9.458,4.329l46.945,54.46L106.918,96.038z M176.116,230.114    l-22.548-46.037l22.701-16.503l36.838,35.644L176.116,230.114z"/>
```
philpl: This is extremely pretty, but the code is waaaaaayyyyy too long :) Remember that this actually shows up and is editable.

components/advanced/tips-and-tricks.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+const TipsAndTricks = () => (
+    <div><h1>Tips And Tricks</h1></div>
```
philpl: To be deleted?

oliverturner: Uh... yeah :)

components/faqs/nesting.js
oliverturner: That's true, but you may recognise it from somewhere ;)
Glenn's CSSConf EU Talk

components/faqs/reverse-selectors.js
oliverturner: Sure, but it only shows up if you scroll down. I'll change if you insist though 😭

pages/docs.json
oliverturner: Huh... TIL acronyms don't take an apostrophe when they lack periods.

components/faqs/nesting.js
philpl: hahaha yea, but we have space here to expand a little on this example

oliverturner: Care to suggest a sufficiently illustrative interpolation?

components/faqs/nesting.js
```diff
@@ -0,0 +1,44 @@
...
+\`;
+render(
+  <ColorChanger href="#">
+    <h2>Hello world!</h2>
+  </ColorChanger>
+)
+`).trim()
+
+const Nesting = () => (
+  <SectionLayout title="Nesting">
```
philpl: @oliverturner Last thing :) Can we change the titles to questions? I'd expect that it feels more natural under an FAQ section

oliverturner: Not really sure that these are questions?! I'd almost suggest reverting calling it "Tips & Tricks" or "Best Practices"?

components/faqs/nesting.js
geelen: I've always liked this example for nesting:

```js
const EqualDivider = styled.div`
  display: flex;
  padding: 1rem;
  ${ props => props.vertical && 'flex-direction: vertical;'  }

  > * {
    flex-grow: 1;

    &:not(:first-child) {
      ${ props => props.vertical ? 'margin-top' : 'margin-left' }: 1rem;
    }
  }
`
```

Adapted from this slide from [this talk](https://www.youtube.com/watch?v=XR6eM_5pAb0):

image

components/faqs/nesting.js
philpl: The problem is, that a "Best Practices" section that I'm thinking of shouldn't be too restrictive, but just include some general rules and recommendations. Otherwise people will think that they can't use it, when they don't agree with some of them.

I wouldn't call it "Tips & Tricks". I've absorbed a lot of them into the other sections and new sub sections. Oftentimes these are so important that they're basics.

The FAQ section is for any question that is not sufficiently addressed by the rest of the docs, and are thus "outstanding questions", in this sense. This is why I'd rephrase it, because the problem is that people would ask these questions anyway on GitHub, Gitter, or Twitter, if we don't answer them.

If we make this section into anything else than an FAQ, we'll lose control over our structure, because it's very tempting to treat such a section as a "Miscellaneous" section, to put everything into it, that takes too long to add somewhere else.

If you think about it, the following sections could also be regarded as "Tips & Tricks", but are integrated into the "narrative" of the Basics and Advanced intros/sections:

- Animations
- Refs
- Security
- Media Templates
- Flow & TypeScript

Even though these are now neatly inside the other sections, they used to be separate. ;)

components/faqs/nesting.js
philpl: This slide shows nicely that you can do basically everything with nesting, that you can do with CSS (obviously). So that's very nice :)

The two interpolations are crucial, since they show that often you can combine nesting + interpolations for more powerful patterns :+1:

The only thing I'm missing is that you should introduce styled components as interpolations for their classname. But you might want to add that to an additional sub-section.

components/faqs/reverse-selectors.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+import { withTheme, ThemeProvider } from 'styled-components'
+
+import SectionLayout from '../SectionLayout'
+import LiveEdit from '../LiveEdit'
+
+const scope = { withTheme, ThemeProvider }
```
philpl: to be deleted? :) seems like you don't need any theming stuff in the LiveEditor here

pages/docs/faqs.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import DocsLayout from '../../components/DocsLayout'
+
+import Nesting from '../../components/faqs/nesting'
+import ReverseSelectors from '../../components/faqs/reverse-selectors'
+
+const APIReference = () => (
```
philpl: Component name's copy-pasta :P

components/faqs/nesting.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import { withTheme, ThemeProvider } from 'styled-components'
+
+import SectionLayout from '../SectionLayout'
+import CodeBlock from '../CodeBlock'
+import LiveEdit from '../LiveEdit'
+
+const scope = { withTheme, ThemeProvider }
```
philpl: Same here. Not needed :)

pages/docs/faqs.js
oliverturner: 😳

components/faqs/reverse-selectors.js
oliverturner: Doh

components/faqs/nesting.js
oliverturner: Oops

components/faqs/nesting.js
oliverturner: > you should introduce styled components as interpolations for their classname

Can you explain what you mean by this?

philpl: E.g. ${Link} where Link is a styled component :)

oliverturner: Ok cool... so would highlighting that very usage in the Reverse Selectors section satisfy?

components/faqs/nesting.js
oliverturner: > If you think about it, the following sections could also be regarded as "Tips & Tricks"

I did think about it, which is why I'm wondering whether these examples wouldn't be better with the same treatment. As much as anything, I'm not sure what the question here is! Is it "can I nest?" or is it broader: "What are the gaps with Sass"?

oliverturner: I'd class co-locating media queries and Reverse Selectors as entirely optional "best practises" fwiw

philpl:
Awesome job :) Thx a lot! Let's see if we can fill it up with more stuff as issues come in

PR #47, opened 2017-06-12
Add missing react/-native imports to example code

craigayre:


components/basics/react-native.js
```diff
@@ -6,6 +6,8 @@ import Note from '../Note';
 import Link from '../Link';
 
 const sample = (`
+import React from 'react';
+import { Text, View } from 'react-native';
```
philpl: This `react-native` import is actually not necessary :wink:

SaraVieira:
@philpl @morajabi Is this PR still valid ?

morajabi:
@SaraVieira I don't think so, and there's no such path anymore, but we need to add that React import here https://github.com/styled-components/styled-components-website/blob/master/sections/basics/react-native.js#L10
```js
import React from 'react'
```

SaraVieira:
closed by https://github.com/styled-components/styled-components-website/pull/125

philpl:
Alright I’ll close this for now. Once I add custom transpilers we could use Babel-standalone and traverse the AST to remove import statements, and add them to all examples

PR #48, opened 2017-06-15
Theme color that goes with the header

siddharthkp:
Use a theme color that goes well with the header.

Before & after:


philpl:
Cheers :)

PR #55, opened 2017-07-07
fix: more consistent example

evgenyrodionov:
Just replaced `css` for `styleTags`, because `sheet.getStyleElement()` is really returning style tags, not plain css.


philpl:
Good point! Thanks :)

PR #57, opened 2017-07-12
Add option to run flow-typed with npx

josephrexme:
flow-typed is an example of a package that'd rarely be used so it might be a good idea to run it through `npx` for those with npm versions that support it

### DIFF
--- a/components/api/flow.js
+++ b/components/api/flow.js
@@ -16,6 +16,10 @@ const flowconfig = (`
 .*/node_modules/styled-components/.*
 `).trim()
 
+const npxinstall = (`
+npx flow-typed install styled-components@<version>
+`).trim()
+
 const Flow = ({ url }) => (
   <SectionLayout title="Flow">
     <p>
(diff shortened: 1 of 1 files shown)


components/api/flow.js
```diff
@@ -35,6 +39,10 @@ const Flow = ({ url }) => (
     <SectionLayout sub title="Installing the definitions">
       <CodeBlock code={install} />
     </SectionLayout>
+  
+    <p>If you are on npm >= 5.2 you could simply use npx</p>
```
philpl: Maybe we should use the `<Note>` component for this and just add a link to the `npx` repo?

josephrexme: Sounds great. But I just looked through npx repo and it doesn't shoot the message quick enough to anyone that has never used `npx`. Someone aware of npx would get it but we may not be saving people without a pre-existing knowledge of what npx is from globally installing flow-typed. Just a thought. If you think having a link is the best way to go I could make that as well

philpl: Ok fair enough :) Maybe we can use this article then? https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b

josephrexme:
@philpl I've made suggested changes. Looks good?

PR #59, opened 2017-07-14
Show how to import keyframes helper

TrevorBurnham:
The Basics page doesn't show `keyframes` being imported, leaving some ambiguity about where that helper comes from. It is `styled.keyframes`? `import 'styled-components/keyframes'`?  I've tried to clarify here.

Ideally the `import` statement would be in the live code sample, but I've settled for putting it in the preceding description.

### DIFF
--- a/components/basics/animations.js
+++ b/components/basics/animations.js
@@ -37,8 +37,8 @@ const Animations = () => (
     </p>
 
     <p>
-      This way, you get all the benefits of using JavaScript, are avoiding name clashes and get your keyframes
-      like always:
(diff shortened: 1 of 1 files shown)


philpl:
Ok, so the text you added adds what is already stated in the paragraph before, and deletes text that tells the user what it means to use it. Instead we can use the `<Note>` component and just say sth along the lines of:

> `keyframes`, like other helpers, is another export of the library. You can import it as follows: `import { keyframes } from 'styled-components'`

Additionally to this, we probably need to write another section in "Getting Started" that speaks about the imports briefly.

TrevorBurnham:
Sounds good to me. 👍 I'll leave it to you to figure out the best approach here.

PR #61, opened 2017-07-18
Add snapshot tests

MicheleBertoli:
I wrote a few tests to try the new version of `jest-styled-components` and find out some edge cases, feel free to merge this if you think it adds value to the project.

screen shot 2017-07-18 at 15 08 01

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,4 @@
 {
-  "presets": [
-    "next/babel"
-  ],
   "plugins": [
     ["babel-plugin-styled-components", {
       "displayName": false,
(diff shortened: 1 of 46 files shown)


philpl:
This is truly amazing! Thanks a lot, Michele :+1: :tada: :100:

cc @mxstbr, do you want to take a quick look as well, before I merge this?

.babelrc
```diff
@@ -1,5 +1,6 @@
 {
   "presets": [
+    "env",
```
philpl: Just noticed this small change :) It shouldn't be necessary (and might actually cause problems with the module transpilation?) https://github.com/zeit/next.js/blob/v3-beta/server/build/babel/preset.js#L16

MicheleBertoli: Unfortunately it's necessary for testing, and removing it makes Jest complain:
> SyntaxError: Unexpected token import

I updated the `.babelrc` following [the example](https://github.com/zeit/next.js/blob/v3-beta/examples/with-jest/.babelrc) on the repo of Next.js, so there's should be no problems.

philpl: Perfect! :)

philpl:
I'm going to go ahead an merge this and will set up Travis later :)

MicheleBertoli:
🙌 Yay, thanks @philpl.

PR #69, opened 2017-08-09
Use fully qualified plugin name in Babel config

chucksellick:
Using just "styled-components" can cause a hard-to-understand error
("Cannot create styled-component for component: [object Object]") if
the plugin hasn't actually been installed, since Babel tries to load
the styled-components module itself instead of the plugin. Fully
qualifying the plugin name ensures that a sensible error is thrown
if the plugin isn't present.


philpl:
Didn't even notice that it's using the shorthand in the examples when I copied them over. Good spot!

PR #72, opened 2017-08-18
update snpashots

SaraVieira:


SaraVieira:
@philpl This fixes the snpashots I forgot to push previously :(

My bad

philpl:
Hm, I upgraded styled-components on the repo and updated the snapshots already and it seemed fine 🤔 don't think this is actually correct in this case then, right?

SaraVieira:
Updated the base to be master :)

philpl:
@SaraVieira I think you have to also install to get styled-components@v2.1.2, because I think the snapshots on master seem to be up to date already. 🤔 💭

SaraVieira:
Oh, then I shall close this 😛
My bad

Issue #64, opened 2017-07-24
Add Tooling Section


MicheleBertoli:
We want to create a Tooling section, where we are going to add:
- `babel-plugin-styled-components` (now under Advanced, see: styled-components/styled-components-website/pull/62)
- `jest-styled-components`

bhough:
Would also like to add 'polished` to this as well.

SaraVieira:
Maybe we should also add the stylelint configuration to the section as well

mxstbr:
This is great! We should definitely move the stylelint processor plus the Babel plugin over, opening new issues for that.

PR #62, opened 2017-07-23
Babel plugin

SaraVieira:
Closes #13 and creates a section in advanced with the babel plugin docs

PR #74, opened 2017-08-24
Add Markdown util and convert all sections

philpl:
I adapted react-markings for our components: https://github.com/Thinkmill/react-markings

It should actually work better for us than converting everything to Gatsby right now, and at the same time will make the task of translating everything a lot smoother probably. Not to mention the connected refactor and redesign.

It should be relatively straightforward, apart from the manual labour that's in translating everything to markdown :(

Edit: Not building with Uglify / build task right now due to an UglifyJS bug

PR #76, opened 2017-08-27
Add Tooling Section 

MicheleBertoli:
Following the discussions #62 in and #64, this PR implements a "Tooling" sections.
It also moves the babel plugin page from "Advanced" to "Tooling".


philpl:
Nice! I really like the idea of having a separate section for tooling! It opens up the possibility of adding a lot more utilities in the future :tada:

I’m working on a new branch that switches all sections over to be written in markdown instead. Do you think we should merge this first and incorporate your changes into it? I’m not quite sure which way around it’s going to be easier, as there’s some miscellaneous changes in both

https://github.com/styled-components/styled-components-website/pull/74

mxstbr:
Yesss this is super awesome!!

We want to integrate the docs for the stylelint processor too, and all the Babel plugin docs, so that in the end a) more people will find the tooling and b) it's all in one place 👌

philpl:
I'll go ahead and merge the section over at #77, which is this PR + the merged back changes :+1: :smile:

MicheleBertoli:
Thank you very much @philpl @mxstbr

Issue #82, opened 2017-08-28
Improve SEO


mxstbr:
I know, I know, who needs SEO anyways? But if you google the site this is what you get:

![screen shot 2017-08-28 at 19 53 27](https://user-images.githubusercontent.com/7525670/29786214-99db384e-8c2a-11e7-97a5-bf0adbbafed3.png)

None of the content, and only the top-level pages are linked. (not even the sub-sections are there) Would much appreciate help so people can find the content on the website more easily!

philpl:
> Diese Seite übersetzen

Subliminal message from Google to push #70 haha

This should be simple enough. Now that we have markdown, it should be possible to extract some page information automatically even? But since the main docs pages contain sub sections none of the content might be relevant.

Maybe the best thing to do is to just list the sub sections in the descriptions?

Both is not that nice, I'd love some input and ideas ❇️

mxstbr:
I think we can probably somehow tell Google that the subsections are pages, linkable with the hashes, I'm just not 100% sure how to do that!

harshmaur:
Guys, why arent we using next/head to define title and descriptions for each page? I know a bunch of other tags as well that are good for SEO purposes.

mxstbr:
> Guys, why arent we using `next/head` to define title and descriptions for each page? I know a bunch of other tags as well that are good for SEO purposes.

PRs very _very_ welcome!!

harshmaur:
@mxstbr Can you give me all the social media pages of styled-components? I am creating the schemas, breadcrumbs etc.

mxstbr:
The only "social media" we have at the moment is Medium, that's all: https://medium.com/styled-components

harshmaur:
Any plans to create other accounts? twitter atleast? or shall I use @mxstbr as the handle?

philpl:
Fixed in #92

@mxstbr Should be sufficient, right?

mxstbr:
Opening because I want to remember to add it to Google Webmaster Tools, let's see what that says we need to do!

harshmaur:
Hey, lets do this. Its just a meta tag it asks you to add.

PR #92, opened 2017-08-29
Should Now improve SEO #82

harshmaur:
Added dynamic description fields as well as other fields like canonical.

### DIFF
--- a/components/DocsLayout.js
+++ b/components/DocsLayout.js
@@ -7,7 +7,7 @@ const DocsLayout = ({ children, title="", description="" }) => (
   <Container>
 
     <Head
-      title={`styled-components ${title}`}
+      title={`styled-components${title ? `: ${title}` : ''}`}
       description={description}>
         <meta name="robots" content="noodp" />
     </Head>
--- a/components/DocsLayout.js
+++ b/components/DocsLayout.js
@@ -1,13 +1,15 @@
 import React from 'react'
-import Head from 'next/head'
+import Head from './SeoHead'
 import Navbar from './Navbar'
 import { Container, Content, Title } from './Layout'
 
-const DocsLayout = ({ children, title }) => (
+const DocsLayout = ({ children, title="", description="" }) => (
   <Container>
-    <Head>
-      <title>styled-components: {title}</title>
-      <meta name="twitter:title" content={`styled-components: ${title}`} />
+
+    <Head
+      title={`styled-components ${title}`}
+      description={description}>
+        <meta name="robots" content="noodp" />
     </Head>
 
     <Navbar />
--- a/components/SeoHead.js
+++ b/components/SeoHead.js
@@ -1,88 +1,61 @@
 import NextHead from 'next/head'
 import { Component } from 'react'
 
-const defaultTitle = 'styled-components'
-const defaultDescription = 'Visual primitives for the component age. Use the best bits of ES6 and CSS to style your apps without stress 💅'
-const defaultImage = 'https://www.styled-components.com/static/atom.png'
-
 // const CreateMarkup = text => ({ __html: text })
 
 
 export default class SeoHead extends Component {
   render() {
-    const props = this.props
+    const {
+      title = "styled-components", 
+      description="Visual primitives for the component age. Use the best bits of ES6 and CSS to style your apps without stress 💅", 
+      image="https://www.styled-components.com/static/atom.png", 
+      children, 
+      canonical, 
+      url = ""
+    } = this.props
+    const properCanonical = canonical || url
     return (
       <NextHead>
         <title>
-          {props.title || defaultTitle}
+          {title}
         </title>
 
-        <meta name="description" content={props.description || defaultDescription} />
+        <meta name="description" content={description} />
 
         {/* Open Graph */}
         <link itemProp="url" href="https://styled-components.com/"/>
-        <meta itemProp="name" content={props.title || defaultTitle} />
-        <meta itemProp="description" content={props.description || defaultDescription} />
+        <meta itemProp="name" content={title} />
+        <meta itemProp="description" content={description} />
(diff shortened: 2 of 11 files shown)


components/SeoHead.js
```diff
@@ -0,0 +1,89 @@
...
+
+        <link rel="canonical" href={`https://www.styled-components.com${props.canonical || props.url || ''}`} />
+
+        {props.children}
+
+        <link rel="shortcut icon" href="/static/atom.png" />
+        <link rel="icon" href="/static/atom.png" />
+        
+        
+        {/*<script
```
philpl: To be deleted? It's great to have, but we can keep this for another day

components/SeoHead.js
```diff
@@ -0,0 +1,89 @@
+import NextHead from 'next/head'
+import { Component } from 'react'
+
+const defaultTitle = 'styled-components'
+const defaultDescription = 'Visual primitives for the component age. Use the best bits of ES6 and CSS to style your apps without stress 💅'
+const defaultImage = 'https://www.styled-components.com/static/atom.png'
+
+// const CreateMarkup = text => ({ __html: text })
```
philpl: To be deleted; see below

components/SeoHead.js
```diff
@@ -0,0 +1,89 @@
+import NextHead from 'next/head'
+import { Component } from 'react'
+
+const defaultTitle = 'styled-components'
```
philpl: Can you use `defaultProps` instead of the OR behaviour below? Should clean things up a little

components/SeoHead.js
```diff
@@ -0,0 +1,89 @@
...
+export default class SeoHead extends Component {
+  render() {
+    const props = this.props
+    return (
+      <NextHead>
+        <title>
+          {props.title || defaultTitle}
+        </title>
+
+        <meta name="description" content={props.description || defaultDescription} />
```
philpl: Personally, I'd prefer destructuring here, since `props.*` is not really useful and `this.props.*` is obviously a tad long.

pages/docs/index.js
```diff
@@ -42,7 +42,7 @@ const SubHeader = styled.h3`
 `
 
 const Documentation = () => (
-  <DocsLayout title="Documentation">
+  <DocsLayout title="Documentation" description="Get the best out of styled components">
```
philpl: This sounds a tad weird. We've previously just used:

> Use the best bits of ES6 and CSS to style your apps without stress 💅

But here I'd go for something really explicit like:

> Learn how to use styled-components and to style your apps without stress

pages/docs/advanced.js
```diff
@@ -11,7 +11,7 @@ import TaggedTemplateLiterals from '../../sections/advanced/tagged-template-lite
 import ServerSideRendering from '../../sections/advanced/server-side-rendering'
 
 const Advanced = () =>
-  <DocsLayout title="Advanced">
+  <DocsLayout title="Advanced" description="Advanced usage of styled components - Theming, refs, Security, Existing CSS, Media Templates, Tagged Template Literals, SSR">
```
philpl: The default spellingzzz of styled-components is lowercase and with a dash: "styled-components";

So sorry to be bothering you with this :smile: But we've previously increased our consistency with it

:x: styled components
:x: Styled Components
:x: StyledComponents
:x: / ✅  StyledComponent (Note: Ok to use to refer to the class but not the library)

pages/index.js
```diff
@@ -184,6 +184,9 @@ class Index extends Component {
   render() {
     return (
       <div>
+        <Head title="styled-components">
```
philpl: Can we use the actual name, `SeoHead`? It's a bit confusing otherwise

components/DocsLayout.js
```diff
@@ -1,13 +1,15 @@
...
 
-const DocsLayout = ({ children, title }) => (
+const DocsLayout = ({ children, title="", description="" }) => (
   <Container>
-    <Head>
-      <title>styled-components: {title}</title>
-      <meta name="twitter:title" content={`styled-components: ${title}`} />
+
+    <Head
+      title={`styled-components ${title}`}
```
philpl: This is changing the separator from `:` to just a space.
Is this actually desired?

cc @mxstbr

components/DocsLayout.js
```diff
@@ -1,13 +1,15 @@
...
+      description={description}>
+        <meta name="robots" content="noodp" />
```
philpl: I'm curious. What's this really specific robots tag for?

components/SeoHead.js
harshmaur: Not really, I just kept it for some day, we might need this. I also have a snippet for breadcrumbs, I did not find it in the UI so preferred not to add it.

philpl: This is what the git history is for :wink: I'd really prefer not to keep commented out code when it's referenced in this PR, in the git history, and can be moved to a gist. Sorry :/

components/DocsLayout.js
harshmaur: Normally google can make up its own descriptions or take it from other websites, if we put this tag, we can explicitly instruct google to use the ones we mention

pages/docs/index.js
harshmaur: Would be great if you can give me the descriptions for all the pages. :)

philpl:
@harshmaur Awesome! Thanks a lot already. :+1: There's only the spelling of the lib's name outstanding.

> The default spellingzzz of styled-components is lowercase and with a dash: "styled-components";

harshmaur:
Done!

philpl:
Awesome! I'll quickly try it out locally before merging :tada:

philpl:
Merged! Thanks a lot for such a quick contribution to solve the issue :heart:

Issue #91, opened 2017-08-29
Add Analytics


mxstbr:
`micro-analytics` is nice and all, but doesn't quite give us what we need and I don't have time to improve it further.

Let's switch to something else. I'd love for the stats to be public, if that's not possible because it doesn't exist let's just use Google Analytics because it's easy?

philpl:
> micro-analytics is nice and all, but doesn't quite give us what we need and I don't have time to improve it further.

It also seems to be a bit brittle and crashed every now and then, I believe... Although it could also just be Zeit's infra not being well suited for it.

I'd be in favour of Google Analytics. It's the obvious choice

harshmaur:
I have added GA in mine. Let me create a PR. :)

harshmaur:
Can I get the Analytics UA Code please?

harshmaur:
Also, I am pretty bad with jest, it is throwing up an error when I am trying to update the snapshot, it seems like it is coming from the library itself. Not sure how I can fix that. https://cl.ly/150H1C3T2v1N

harshmaur:
I am actually done with the changes, Just need to solve the Jest issue.

mxstbr:
It seems like `react-ga` can't be server-side rendered?

harshmaur:
I am not sure. It works in my project. I don't use tests though. 

Secondly GA should be executed on frontend only right?

mxstbr:
Yeah it should only happen on the frontend, otherwise it'll seem like we have 2x the pageviews! I created a property on Google Analytics, the tracking ID is `UA-105613776-1` :blush:

mxstbr:
This thread has a long discussion about this exact feature:

harshmaur:
I've used the same thing with modifications. How about I remove the test and make a PR then you can have the tests pass?

mxstbr:
> How about I remove the test and make a PR then you can have the tests pass?

Just submit the PR with the failing tests so we can take a look at it. Thanks!!

PR #95, opened 2017-08-29
Added React-GA for Google Analytics #91

harshmaur:


mxstbr:
Yeah so the issue here is that `react-ga` can't be server-side rendered. You should be able to add a `if (window !== undefined)` clause before the `initGa` thing, ref: https://github.com/react-ga/react-ga/issues/98

That should fix it.

harshmaur:
Cant get it to work :(

philpl:
@mxstbr @harshmaur Have you tried it with `typeof window === 'object'` instead of comparison to `undefined` directly? Otherwise this will throw a referenceerror.

mxstbr:
The test is still failing in `react-ga` though, not our code 😕

philpl:
@mxstbr That's probably because the test runs in a browser-like environment (jsdom), but not in a real one. So the node is searching for probably doesn't exist there 😢

mxstbr:
Yeah but that means it would also fail if it was server-side rendered.

mxstbr:
Uhh, what just happened? I didn't close this at all, I was only trying to fix the test! 😭  Where did the branch go?!

mxstbr:
I'm so sorry @harshmaur I just did something very weird, would you mind submitting that PR again? Super sorry

harshmaur:
I dont know what just happened, it seems like a new PR got created with the changes you made, When I am trying to compare for the changes, I do not see any and I am not able to make a PR 🤔

harshmaur:
I think you can reopen the PR?

mxstbr:
I can't reopen it either, the button is greyed out. I have no clue what just happened! Can you push your changes again?

PR #108, opened 2017-09-05
Clarify className usage on styled(Component)

matan129:
Just added a little clarification there - it's very easy to miss in the docs. 
My team and I stumbled on this subject for couple of days.

By the way - awesome library! :trophy:

### DIFF
--- a/sections/basics/styling-any-components.js
+++ b/sections/basics/styling-any-components.js
@@ -3,8 +3,10 @@ import md from 'components/md'
 const StylingAnyComponents = () => md`
   ## Styling any components
 
-  The \`styled\` method works perfectly on all of your own or any third-party
-  components as well, as long as they pass the \`className\` prop to their rendered sub-components.
+  The \`styled\` method works perfectly on all of your own or any third-party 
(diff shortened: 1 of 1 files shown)


sections/basics/styling-any-components.js
```diff
@@ -4,7 +4,7 @@ const StylingAnyComponents = () => md`
   ## Styling any components
 
   The \`styled\` method works perfectly on all of your own or any third-party
-  components as well, as long as they're accepting the \`className\` prop.
+  components as well, as long as they pass the \`className\` prop to their rendered sub-components.
```
mxstbr: This is not entirely accurate, the className just has to be attached to a DOM node!

matan129: Okay. Just to be sure I understand - given that component A renders component B which in turn renders an actual DOM node (say, a `div`) - it's *not* enough to styled(A) to pass the `className` to B, but B also has to pass it down to the `div`?

matan129:
Just wondering - when will this be deployed?

PR #110, opened 2017-09-08
feat(company-logos): add huffpost

cogell:


mxstbr:
Do you work at HuffPo/do you have a source on what they use it for?

cogell:
Yea I'm an engineer at HuffPost. You can check grep for my name "Cedric Cogell" here http://www.huffingtonpost.com/static/about-us

We are currently using styled-components for internal web apps. We're looking forward to having styled-components generate style sheets at build time to improve cold cache, first view performance.  And maybe some day we can use css-in-js on our front page tools :)

cogell:
❤️  thanks!

Issue #80, opened 2017-08-28
Add "Guides" section


mxstbr:
We could use a more tutorial-like guides section for some longer topics.

- Server-side rendering with Express (including setting up a build process?)
- Setting up a codebase for success (stylelint + prettier + folder structure?)
- Preact + styled-components
- Inferno + styled-components
- Other ideas?

philpl:
I'd love to also mix these up with some Youtube videos or links to other resources. We have some great content on Medium for example, so it'd make sense not to try to replace them, but to complement and integrate them instead :+1:

mxstbr:
Yep, we could host the Guides on Medium and talks on YouTube and then have a "Resources" section? I'm just a bit worried that that'll end up being just like a link dump.

SaraVieira:
@mxstbr @philpl I could do a video with the second one if you guys want , use react-create-app and set it up

mxstbr:
That sounds awesome @SaraVieira!

SaraVieira:
@mxstbr @philpl I have been doing preact screencasts for the last couple of days so you can get a sense of how I do them you can see them here:
https://www.youtube.com/playlist?list=PL1O2bxLvC8EPw09PKUl3smGrujiZbQFG9

Not public yet but json already approved them 🎉

PR #114, opened 2017-09-25
Add link to video intro with Jack & Glen.

jackfranklin:
Thought I'd add a link to this video from the docs.

This might not be the best place for it - I'm happy to move it as you wish.

Jack :)


philpl:
I'd love to add it, but I'm not sure whether that's the right section for it :smile:

We actually have a "Guides" section planned that'd be perfect for it. The problem is that it doesn't exist yet 😢 

https://github.com/styled-components/styled-components-website/issues/80

PR #120, opened 2017-09-28
Add styled-theming

### LINKED ISSUE #118: Add styled-theming docs to Tooling section
We should move the https://github.com/styled-components/styled-theming documentation to the website too, likely under the tooling section? Would love more folks to use that package.

/cc @thejameskyle for good measure.

thejameskyle:
Resolves #118


philpl:
@thejameskyle Awesome! Sorry for not pushing for this ourselves though :( We should make sure to link to this from our theming section in the future :wink: so it becomes a well-known best practice

PR #121, opened 2017-09-28
Update configuration for SSR with Next.js

pauldariye:
Update the docs for server-side-rendering with Next.js and add reference to an example from the Next.js repo.

### DIFF
--- a/sections/advanced/server-side-rendering.js
+++ b/sections/advanced/server-side-rendering.js
@@ -56,24 +56,26 @@ const ServerSideRendering = () => md`
   import { ServerStyleSheet } from 'styled-components'
 
   export default class MyDocument extends Document {
-    render() {
+    static getInitialProps ({ renderPage }) {
       const sheet = new ServerStyleSheet()
-      const main = sheet.collectStyles(<Main />)
+      const page = renderPage(App => props => sheet.collectStyles(<App {...props} />))
       const styleTags = sheet.getStyleElement()
+      return {
+        ...page, 
+        styleTags
+      }
+    }
 
+    render() {
       return (
         <html>
           <Head>
             {/* ... */}
 
-            {styleTags}
+            {this.props.styleTags}
           </Head>
-
           <body>
-            <div className="root">
-              {main}
-            </div>
-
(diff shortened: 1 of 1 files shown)


sections/advanced/server-side-rendering.js
```diff
@@ -56,34 +56,35 @@ const ServerSideRendering = () => md`
...
     }
   }
   \`\`\`
-
-  Here we're wrapping the \`Main\` component, which contains the Next.js app, and are using this
-  to extract the styles on the server side.
+ 
+  > Refer to [Next.js example `with-styled-components`](https://github.com/zeit/next.js/tree/master/examples/with-styled-components)
 
   > This is unfortunately only a workaround! It will accumulate rules over time, so you will need
```
philpl: We should be able to remove this note now :smile:

sections/advanced/server-side-rendering.js
```diff
@@ -56,34 +56,35 @@ const ServerSideRendering = () => md`
...
         </html>
       )
     }
   }
   \`\`\`
-
-  Here we're wrapping the \`Main\` component, which contains the Next.js app, and are using this
-  to extract the styles on the server side.
+ 
+  > Refer to [Next.js example `with-styled-components`](https://github.com/zeit/next.js/tree/master/examples/with-styled-components)
```
philpl: * the

And this won't need to be a note (Remove `> `)

Also, maybe we can make this a little nicer? "Refer to [our example] in the next.js repo for an up-to-date usage example."

pauldariye: Thanks, @philpl. I've updated the language.

sections/advanced/server-side-rendering.js
pauldariye: Removed

sections/advanced/server-side-rendering.js
pauldariye: I've removed the leading `>`. Should be good now.

philpl:
Thanks a lot for fixing this up :tada:

PR #122, opened 2017-10-01
Implement Google Analytics

### LINKED ISSUE #91: Add Analytics

mxstbr:
Closes #97, closes #91

### DIFF
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -176,6 +176,16 @@ export default class MyDocument extends Document {
           <style dangerouslySetInnerHTML={{ __html: resetStyles }} />
 
           {styles}
+
+          <script dangerouslySetInnerHTML={{ __html: `
+            !function(s,t,y,l,e,d,c){s.GoogleAnalyticsObject=y;s[y]||(s[y]=function(){
+            (s[y].q=s[y].q||[]).push(arguments)});s[y].l=+new Date;d=t.createElement(l);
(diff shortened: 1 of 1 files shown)


harshmaur:
This kind of insertion works now with react and next?

mxstbr:
Worked locally for me 🤷‍♂️ Any better ideas?

harshmaur:
Did it record all the pages as well? What I know is that way of implementation will only record the landing page and subsequent page navigation wont be recorded.

mxstbr:
Could be, I didn't test it very thorougly. PR with a better implementation very very welcome!

harshmaur:
Hey I did that twice? 🤔 The ones you closed.

mxstbr:
Ohhhh yeah, but those weren't working right? I took that branch and tried to fix it but gave up and did this instead because I thought it'd work that way 😅

harshmaur:
Oh okay. Its the test which is failing, the tracking works fine.

mxstbr:
Yeah but I mean, that's not exactly great :wink: If you can fix that I'd be a lot more happy shipping your version!

Issue #123, opened 2017-10-02
Link to awesome-styled-components list


mxstbr:
styled-components has such a great community behind it and there's so many modules built with it! We should link to https://github.com/styled-components/awesome-styled-components from the website to make sure people can discover them.

SaraVieira:
@mxstbr Where do you think this should be linked ?

mxstbr:
No idea!

morajabi:
A community link in the navbar?

mxstbr:
"Community" should, if anything, like to Spectrum, but what about "ecosystem"? We need to figure out what to put in the navbar anyway, see #115, so we can finally ship your awesome updates @morajabi!

morajabi:
@mxstbr You're totally right, Max! I like "ecosystem". Also, it sounds like a proper link to put in out empty navbar :)

SaraVieira:
And should that page simply have the contents of the readme in awesome styled components?

mxstbr:
We either pull it in via the API or something (so we don't have to keep two lists up to date), or we just like to the repo.

philpl:
We have a task tracking this: https://github.com/styled-components/styled-components-website/issues/99

I think having a generator API that pulls the readme from the awesome repo and transforms it would be optimal.

morajabi:
Yeah, that's right. @mxstbr himself opened that too :)

SaraVieira:
Yellow !

I'm giving this a try with the `md` component but I'm having some hardship :(

Is this the correct way to use it ?

```
import React, { Component } from 'react'
import DocsLayout from '../components/DocsLayout'
import { getReadme } from '../utils/githubApi'
import md from '../components/md'

class Ecosystem extends Component {
  constructor() {
    super()
    this.state = {
      loading: true,
      readme: null
    }
  }

  componentWillMount() {
    getReadme().then(readme =>
        this.setState({
          readme,
          loading: false
        })
    )
  }

  render() {
    return (
      <DocsLayout
        title="Ecosystem"
        description="Ecosystem of styled-components">
        {!this.state.loading && md(this.state.readme)}
      </DocsLayout>
    )
  }
}

export default Ecosystem
```

I will place the state stuff into a readme component but just wanted to get it working first

@mxstbr @morajabi @philpl

morajabi:
[...]

PR #124, opened 2017-10-03
fix: support calling md as a normal function too

morajabi:
As discussed in #123 I added a type check for the `strings` variable


philpl:
Awesome! ⭐️

PR #128, opened 2017-10-04
Integrate Google tag manager

mxstbr:
My previous Google Analytics implementation wasn't working, only firing
on first visit. Google Tag Manager has the power to listen to history
changes and fire analytics events when they happen, so I've configured
Google Tag manager to do exactly that.

This means our analytics setup is now working perfectly and tracking
pageviews as it should.

### DIFF
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -177,18 +177,11 @@ export default class MyDocument extends Document {
 
           {styles}
 
-          <script dangerouslySetInnerHTML={{ __html: `
-            !function(s,t,y,l,e,d,c){s.GoogleAnalyticsObject=y;s[y]||(s[y]=function(){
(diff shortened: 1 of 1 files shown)


mxstbr:
I was following this guide for the google tag manager setup, just in case you're wondering: https://www.pmg.com/blog/tracking-single-page-web-apps-google-tag-manager-analytics/

philpl:
Looks good but I really seem to know nothing about this... 😅

mxstbr:
Yeah I did not know what Google Tag Manager was before doing this, but if it works it works 🤷‍♂️

SaraVieira:
@mxstbr @philpl I have worked with google tag manager in the past , it was painful but if you want to add any events let me know

Issue #130, opened 2017-10-05
Improve inline code styling


mxstbr:
I think our current inline code styling isn't brilliant:

![screen shot 2017-10-05 at 11 32 29 am](https://user-images.githubusercontent.com/7525670/31220537-039ce800-a9c1-11e7-86c8-f75e883a9924.png)

It sticks out of the text like a sore thumb. I much more like GitHubs treatment of `inline code`, but I feel like we just gotta find something that's less obvious and flows better when reading.

morajabi:
Totally agree with this. Can you tell me more about your preference so I can do some sketches?

mxstbr:
No preferences or sketches, happy to review a couple ideas!

morajabi:
OK, let's start with the problems with "current style" and the "Github style".

PR #132, opened 2017-10-05
Update inline-code style

Jinksi:
#130 
Based upon Github `inline code` style.

### DIFF
--- a/components/Code.js
+++ b/components/Code.js
@@ -12,8 +12,8 @@ const Code = styled.span`
   font-weight: 300;
   padding: 0.15em 0;
 
-  &:before,
-  &:after {
+  &::before,
+  &::after {
     letter-spacing: -0.2em;
     content: "\00a0";
   }
--- a/components/Code.js
+++ b/components/Code.js
@@ -1,22 +1,15 @@
 import styled from 'styled-components'
 import { monospace } from '../utils/fonts'
-import { paleGrey } from '../utils/colors'
 
 import '../utils/prismTemplateString'
 
 const Code = styled.span`
-  background: ${paleGrey};
-  border-radius: 3px;
+  display: inline-block;
   font-family: ${monospace};
-  font-size: 85%;
+  font-size: 90%;
   font-weight: 300;
-  padding: 0.15em 0;
-
-  &::before,
-  &::after {
-    letter-spacing: -0.2em;
-    content: "\00a0";
-  }
+  padding: 0 0.05em 0.1em;
+  vertical-align: bottom;
 `
 
 export default Code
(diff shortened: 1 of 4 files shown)


mxstbr:
Screenshot?

Jinksi:
Screenshot:  
screen shot 2017-10-05 at 8 44 23 pm

SaraVieira:
Deployed version: https://styled-components-docs-wbvnnwqfmx.now.sh

SaraVieira:
@Jinksi You need to update the snapshots 😔

mxstbr:
Hmm, looks better but I feel like the background thing doesn't work super well with our typography. The code font is so different from the normal font, do we even need the background? What does it look like with that removed?

Jinksi:
screen shot 2017-10-05 at 8 51 16 pm

SaraVieira:
what do you guys think of changing the font to https://fonts.google.com/specimen/Roboto+Mono ?

mxstbr:
Operator Mono is kinda part of the styled-components identity, I'd hate to remove it.

philpl:
@mxstbr We actually removed the background-box from inline code on purpose after a recommendation from Andrey

SaraVieira:
I think just adding the spacing helped a lot :)

SaraVieira:
@Jinksi One test is still failing 😔

morajabi:
But this much of space and light font makes it like a separator.

philpl:
Small nit, it's a little off centre (Chrome screenshot), and on iOS (not the screenshot) the baseline is a little off

screen shot 2017-10-05 at 12 58 09

The previous implementation we had put the text on the same baseline and used margin+padding (not just padding) to keep the background "floating" in the centre behind the text

mxstbr:
I like it more without the backgroud color but the space on the side needs to be adapted to be the same as an actual space.

Jinksi:
@SaraVieira Jest is throwing `SyntaxError: Unexpected token import` on `/test/components/Code.spec.js` when I run `yarn jest -u`. Any ideas? 😕

Jinksi:
Increase font-size, remove background & vertically align:
screen shot 2017-10-06 at 6 49 36 am

morajabi:
@Jinksi You need to run `yarn jest -- -u` to pass the `-u` arg down to the jest not yarn itself. Also, If it doesn't work either, try removing `node_modules` and reinstalling with `yarn`.

Jinksi:
😅  Sorted it, some cache issue with jest. I cloned a fresh copy and snapshots updated ok.

mxstbr:
Thanks @Jinksi!!

Jinksi:
@mxstbr Cheers! ✌️

PR #133, opened 2017-10-06
Unblock first render by moving font link

mxstbr:
Loading Operator Mono in the `<head>` meant we were blocking the first render by quite a long time. The font had to be completely loaded before the browser could construct the CSSOM, meaning until then nothing was happening making our perceived loading time a lot slower than it needed to be.

By moving the fonts `.css` file at the bottom of the `<body>` we have a first render before that font is loaded, making perceived performance a lot faster.

This does give us a FOUT (flash of unstyled text), BUT since it's only the font we use for code it's unnoticable and/or very small.

PageSpeed Insights before:

screen shot 2017-10-06 at 10 00 55 am

PageSpeed Insights after:

screen shot 2017-10-06 at 10 01 11 am

Check out the FOUT on your own machine by going to this deploy: https://styled-components-docs-qldaglwxhk.now.sh

### DIFF
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -176,8 +176,6 @@ export default class MyDocument extends Document {
           <meta name="theme-color" content="#da936a" />
           <meta name="author" content="styled-components" />
 
-          {/* cloud.typography */}
-          <link rel="stylesheet" type="text/css" href="https://cloud.typography.com/7039052/7606172/css/fonts.css" />
           <style dangerouslySetInnerHTML={{ __html: resetStyles }} />
 
           {styles}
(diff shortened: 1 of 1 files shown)


mxstbr:
WebPageTest before:

screen shot 2017-10-06 at 10 08 53 am

WebPageTest after:

screen shot 2017-10-06 at 10 11 00 am

> Note: This is on a Moto G4 in Dulles, VA on a "Mobile 3G" connection
>
> Note: The TTFB is a second lower because the server is under less load, but even if you add that second to the other times we're still faster to show up on a users screen with this change.

PR #135, opened 2017-10-06
Added Stylelint note about --fix option

emilgoldsmith:
Also fixed a few stray NOTE's (as I noticed that the markdown renderer automatically adds these)


emilgoldsmith:
No problem!

Just rebased to update the changes from the recent PR merge, so if you could just quickly re-approve that'd be great @mxstbr !

Issue #131, opened 2017-10-05
Fix Ecosystem sidebar to use the page headings


morajabi:
@philpl 
> The navbar and sidebar are pretty static and unconfigurable at the moment 😢, However, we'll need to include the navbar on the homepage eventually, and we'll need to grab all the headers (## ...) and generate an alternative sidebar for this page.

We need to grab all headings from the Ecosystem page and build the proper sidebar for this specific page.

chinesedfan:
@morajabi Already done by #143?

mxstbr:
Good catch @chinesedfan!

morajabi:
Oh, it feels good looking how much we've done 😊

PR #138, opened 2017-10-07
[WIP] Add Navbar to homepage

mxstbr:
This is a WIP idea, /cc @morajabi. The idea is that the navbar background is transparent when the page is at the top, and it turns into it's normal pink once the user scrolls:

screen shot 2017-10-07 at 1 22 15 pm

screen shot 2017-10-07 at 1 26 20 pm

It's already fully mobile compat and just doesn't show the sidebar on the homepage. (I added a `showSideNav` prop)

> Note: We should probably enable that prop for the Ecosystem page until we get around to creating a more comprehensive, auto-generated SidebarNav?

This is WIP because even with debouncing scrolling is janky, and the transition needs to be reworked to use opacity rather than background to be smoother too. I don't have more time at the moment, so I'd be happy if somebody could take this branch and finish it off!


morajabi:
Hmm actually @philpl and I have discussed it and we agreed to first refactor sidebar #131 for the Ecosystem page and then get to this. Anyway, it's not a big deal :)

mxstbr:
That seems like more effort and I had a bit of time so I figured might as well get this done :wink:

morajabi:
@mxstbr Great! So is background transition that bad? Also take a look at this https://github.com/brigade/react-waypoint

morajabi:
Did you see the `react-waypoint`?

morajabi:
The problem with the scroll event handler is, it's invoked when I stop scrolling.

mxstbr:
Niceee, much better-thanks @morajabi!

WDYT @philpl?

philpl:
Looks amazing! Let’s ship it 🦋

Issue #136, opened 2017-10-06
innerRef example uses `this` where `this` is the window object


sk22:
In https://www.styled-components.com/docs/advanced#refs, the code says this:

```javascript
const Form = () => (
  <Input
    placeholder="Hover here..."
    innerRef={x => this.input = x}
    onMouseEnter={() => this.input.focus()}
  />
);
```

Since the `Form` component is written as an arrow function, I got confused of the use of `this`, as there's no `this` the `innerRef` and `onMouseEnter` props could use but the global scope. And indeed, after hovering over the text box, `window.input` contains the reference to the input element.

Please change that, because it's bad practice to store variables in the global scope.

sk22:
The example here does it better: https://github.com/styled-components/styled-components/blob/master/docs/tips-and-tricks.md#refs-to-dom-nodes

mxstbr:
PR to clean this up very welcome!

glhrmv:
I'll work on this.

glhrmv:
I've submitted #139. Couldn't extend just `Component` so I hope this is fine.

mxstbr:
Thanks @glhrmv!

PR #139, opened 2017-10-07
Update docs ref example so as to not promote bad practices

glhrmv:
Reference issue: #136

### DIFF
--- a/sections/advanced/refs.js
+++ b/sections/advanced/refs.js
@@ -25,13 +25,17 @@ const Refs = () => md`
     border-radius: 3px;
   \`;
 
-  const Form = () => (
-    <Input
-      placeholder="Hover here..."
-      innerRef={x => this.input = x}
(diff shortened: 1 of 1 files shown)


sections/advanced/refs.js
```diff
@@ -25,13 +25,17 @@ const Refs = () => md`
...
-      innerRef={x => this.input = x}
-      onMouseEnter={() => this.input.focus()}
-    />
-  );
+  class Form extends React.Component {
+    render() {
+      return (
+        <Input
+          placeholder="Hover here..."
+          innerRef={x => this.input = x}
```
philpl: Thanks for fixing this up; while you’re add it can you add a function body around the assignment here, since that’s closer to prettier output. Sorry but a lot of people might feel it being off otherwise

glhrmv:
Is this better? (I didn't mean to "dismiss" anything, not sure how that happened)

PR #142, opened 2017-10-14
Fix video aspect ratio

ink8bit:
# Video from `motivation` page

Fixed an issue with a broken layout of the motivation video.

## Steps to reproduce
1. Go to https://www.styled-components.com/docs/basics#motivation
2. Switch to a mobile device in dev tools

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,7 +21,7 @@
   },
   "env": {
     "mocha": true,
-    "browser": "true"
+    "browser": true
   },
   "parserOptions": {
     "ecmaVersion": 7,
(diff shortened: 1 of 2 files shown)


ink8bit:
> Just one tiny request, could we make it so it doesn't grow to full-width on desktop?

Of course we can! 

@mxstbr  Please, check this commit https://github.com/styled-components/styled-components-website/pull/142/commits/2d07f9ca6814485af389e3015efb7e14a339290b

morajabi:
Shouldn't we use our `mobile` media helper? I think it's in `utils/media`. If it's a special media width, then there's no problem.

ink8bit:
Added a `media` helper https://github.com/styled-components/styled-components-website/pull/142/commits/d4e91e93b527a9da09092be5a3d0083ae9cda000

But I guess that it is `phone`-helper.

morajabi:
@ink8bit Right I hadn't seen that :) It's better now, Thanks!

Issue #105, opened 2017-09-04
Can't link to headings on API page


mxstbr:
One can't link to headings on the API page. They are `<code>` blocks and our slugifier obviously doesn't know what to do with that:

![screen shot 2017-09-04 at 11 03 43 am](https://user-images.githubusercontent.com/7525670/30019399-c369cd48-9160-11e7-8800-73a05f436e9b.png)

We should get that fixed as it's pretty annoying!

philpl:
@mxstbr hm I thought I fixed the stringifier. I’ll take a look this afternoon

morajabi:
Just discovered there's another problem when there's a code block in the title:
This
![screen shot 2017-09-14 at 5 06 06 pm](https://user-images.githubusercontent.com/12202757/30430038-14b307bc-996f-11e7-8090-cd3e3c6a1267.png)
Does not match the link in the home page:
![screen shot 2017-09-14 at 5 07 14 pm](https://user-images.githubusercontent.com/12202757/30430069-3af83f50-996f-11e7-8061-440e2a620467.png)

morajabi:
@mxstbr BTW, I don't see any problems there:
![screen shot 2017-09-14 at 5 06 06 pm](https://user-images.githubusercontent.com/12202757/30430433-73ff7b28-9970-11e7-825f-0a169c8b6600.png)

morajabi:
@philpl Did you see this and #112?

philpl:
This should now be fixed in https://github.com/styled-components/styled-components-website/pull/153

The original bug in this issue was fixed a while ago, but the slugifier was still behaving incorrectly.

PR #143, opened 2017-10-16
Ecosystem sidebar (+ SidebarMenus refactor for future special sidebars?)

morajabi:
Hey! Finally got the time finish this. 
I made a base for future sidebars. Current base only has the minimum feature, if we needed to add something later, it's achievable with some little edit to `SidebarMenus` components.

For now, I'm parsing the ToC of the awesome-styled-components repo by HAND and it *seems* dirty but works perfectly. if you have any ideas share them with me!

https://styled-components-docs-fapuytgaku.now.sh

BTW, @philpl we need to check that stringifier.


mxstbr:
Ref issue for the stringifier: #105

morajabi:
@mxstbr With this version, I managed to not have conflicts with the stringifier issues, but for further development, we need to fix 2 things regarding the hash links, first the strigifier then the problem of external links to the website with hashtags. It seems it's something with NextJS client-side rendering (rehydrating) causing the issue. I mentioned these in #105 but we were busy with other things at that time.

morajabi:
@philpl

morajabi:
@philpl Did you see this? (Sorry, I thought maybe you missed the notification 🙏)

philpl:
@morajabi oh, sorry for not catching up with this sooner! I’ll check out the branch and take a look this evening

morajabi:
Sorry again for that reminder! I feel bad :)

philpl:
@morajabi finally got around to take a look at it (sorry, again)

One nit, when I click the section title, it doesn't jump to it but instead just unfolds the subsections. This applies to "Built with styled-components" for example; click it, expands, clicking the subsection jumps to one. But clicking the section itself doesn't.

One small thing, that we don't have to fix here, this is a little sluggish: https://github.com/styled-components/styled-components-website/blob/master/components/WithIsScrolled.js#L7

Instead of this you can guard the `setState` by an if statement here: https://github.com/styled-components/styled-components-website/blob/master/components/WithIsScrolled.js#L30

This would check the previous state and the new state of `isScrolled`. Since then `setState` would not be called all the time (which incurs a cost), you don't have to debounce. Then the background transition can be sped up and voila, the navbar should feel much more responsive in its transition.

morajabi:
@philpl I'll tell you more, but one thing I use PureComponent so it generates `shouldComponentUpdate` for me.

philpl:
@morajabi Yea, regarding `PureComponent`. You're still paying the cost of that reference equality check, so it's always best not to call `setState` in the first place. Then you can also turn it into a normal `Component`.

> Right, I've seen it and it seems to be a bug with NextJS or the stringifier. I'll take a look but I thought it's also acceptable to just fold/unfold the subsection.

I'll fix the stringifier, but it only fails on inline code, right? Anyway, I'll get #105 fixed

morajabi:
@philpl I have a problem. I rely on `window.location.href` to check whether a sidebar folder is open or not: 
```js
function isFolderOpen(currentHref, { pathname, title, sections }) {
  return (
    sections.reduce((sum, v) => (
      sum || window.location.href.endsWith(getSectionPath(pathname, v.title))
    ), false) ||
    window.location.href.endsWith(pathname || '#' + titleToDash(title))
  )
}
```

Click on a `SectionTitle` should change the URL and then by the function above, we'll open the folder based on the URL. The problem is when the URL changes and `Folder` receives new props `window.location` has not changed yet! Does Next router expose something for this? I couldn't find one.

philpl:
@morajabi I pushed a PR to fix the `WithIsScrolled` responsiveness. Looks a little better to me https://github.com/styled-components/styled-components-website/pull/154

> Does Next router expose something for this? I couldn't find one.

Yes! You can import the router and attach a listener to it:

```js
import Router from 'next/router'

Router.onRouteChangeStart = url => {
  console.log('App is changing to: ', url)
}
```

https://github.com/zeit/next.js/#router-events

morajabi:
@philpl btw regarding that navbar I made the changes but forgot to push hahaha. Although your version did some cool refactoring in `render` too. 

And I'll use that and see how it behaves.

components/Nav/SidebarMenus.js
```diff
@@ -101,4 +103,66 @@ const SidebarMenu = ({ onRouteChange }) => (
...
+function getSectionPath(parentPathname, title) {
+  return `${parentPathname || ''}#${titleToDash(title)}`
+}
+
+function isFolderOpen(currentHref, { pathname, title, sections }) {
+  return (
+    sections.reduce((sum, v) => (
+      sum || window.location.href.endsWith(getSectionPath(pathname, v.title))
+    ), false) ||
+    window.location.href.endsWith(pathname || '#' + titleToDash(title))
```
ink8bit: Hi, @morajabi! Maybe it's better to use a template string in such kind of places in this PR?

morajabi: Actually I used template strings in this file but I think adding a single # to the beginning of another string is simple enough to be done this way. 
`` `#${titleToDash(title)}` `` vs `` '#' + titleToDash(title) ``

philpl:
@morajabi Shall we merge this without a subsection-fold fix, since that doesn't seem to be doable as of now? We can still add it later

mxstbr:

mxstbr:
Btw @philpl even though this shouldn't have sent here this is the message that shows when a PR is merged and the person is invited ^

Dunno why that was sent here though since @morajabi is already a member of styled-components 🤷‍♂️ /cc @orta

morajabi:
Double invited! Whoop whoop! 😍

PR #144, opened 2017-10-19
Included note re: babel-plugin-styled-components

cezarneaga:


morajabi:
Probably we need to update the whole section since the [with styled-components example](https://github.com/zeit/next.js/blob/master/examples/with-styled-components/pages/_document.js) in the Next repo is now exactly the same as[ the snippet in this section.](https://github.com/cezarneaga/styled-components-website/blob/7099ded5fd16ce859dde2b1cf1ba36b35c95f10b/sections/advanced/server-side-rendering.js#L49)

cezarneaga:
Agree. I spoke to @mxstbr and he mentioned the website needs to catch up. 
I came back to next.js with styled components after a while and i remembered there were issues before v2. i tried website docs first, but that didnt work, obviously. :) tried next.js example and didnt work because i hadn't look into the package.json to see the babel plugin. hence my PR

morajabi:
@cezarneaga Good job! Yea, it's definitely necessary.

cezarneaga:
was just curious if we should add a comment about putting this in your`.babelrc`
```
"plugins": [
    [
      "styled-components",
      { "ssr": true, "displayName": true, "preprocess": false }
    ]
  ]
```

mxstbr:
I would just link to the Next example as the canonical source for the code, and add a note on our website that you have to use the Babel plugin?

morajabi:
@mxstbr Fair enough.

Issue #86, opened 2017-08-28
Translation tooling


mxstbr:
With the volunteers in #70 and the changes in #74 the content is now in a markdown string and we should now be able to start translating the webpage.

React uses https://crowdin.com, which seems pretty good in terms of the workflow for translators and they offer a free unlimited account to OSS projects. What I'm not sure about is the workflow from the development side; how are changes going to get on the website, which format do the strings need to have, etc.

On top of that I'm not sure how we'd make that happen across the app, how we'd structure the repo, etc.

Any investigations/experiences here very helpful.

victorhqc:
I'm gonna post a PR soon about this. Gonna be using [this](https://github.com/i18next/react-i18next) and dummy translations just as a start point.

mxstbr:
Awesome, can't wait!

el1f:
Hey @kitten! I checked out #148 and it seems like a lot of work had been done but it was closed due to the absence of a clear focus on what the translation system would need to achieve structure-wise. Are we now in a clearer place to reopen the conversation regarding this extremely interesting feature?

PR #148, opened 2017-10-25
Web Translations

victorhqc:
## Introduction
This Pull Request can help as a starting point to start translating this website. I chose [react-i18n](https://github.com/i18next/react-i18next) as base as it has a pretty good documentation and solid API.

Spanish

### What changed
* added i18n dependency and added basic translations for home page as a proof of concept
* fixed tests in Mac OSX. For some reason test kept failing until I added `'use babel'`
* Update Readme to install dependencies with `yarn` if dependencies are installed with `npm` tests fail and dependencies change.

### What's missing
* Need to get the translations from an external tool like [https://crowdin.com/](https://crowdin.com/)

Related issue https://github.com/styled-components/styled-components-website/issues/86


victorhqc:
If this is merged I'll complete the migration to have all the English language extracted out of the main code. And then more people can join in to start translating more languages (I can help with spanish 😄)

victorhqc:
That is handled by the [Language detector](https://github.com/i18next/i18next-browser-languageDetector) I'll add Spanish translations for home page and see what happens. It should be detected by the `navigator`. Checking

victorhqc:
Confirmed that spanish works
Spanish

Tested english and still was good
English

At the moment only changing the language to the browser triggers this. But we can add a button to choose the desired language

victorhqc:
The server doesn't need to do anything to support the languages. Everything thing is called in the components.

mxstbr:
> The server doesn't need to do anything to support the languages.

The server should server-side render the language that the browser passes via the relevant headers, otherwise everybody is going to get the english one and then only when the client-side js kicks in will it change to their home language?

philpl:
@victorhqc seems like a great start 🛫 I’m not too aware of all the approaches of React + I18N, but have mostly used React Intl before. Are there any pros/cons of this lib compared to that?

victorhqc:
> The server should server-side render the language that the browser passes via the relevant headers, otherwise everybody is going to get the english one and then only when the client-side js kicks in will it change to their home language?

Yeah, it is detected by headers in request. I guess we could implement also some `nginx` rules so we have `styled-components.com/de` `styled-components.com/es` etc. But the default option is just headers and manually changing language if user wants to.

victorhqc:
>@victorhqc seems like a great start 🛫 I’m not too aware of all the approaches of React + I18N, but have mostly used React Intl before. Are there any pros/cons of this lib compared to that?

@philpl As far as I can see, *React Intl* offers a more simple way to achieve the translations. So the pro for *React Intl* is simplicity. However, *i18n* has a more complete and bigger API, as well as a more dynamic ways to use it. Not just using *HOC* it also offers use *dynamic render*. As well as different ways to get the translations. Either using `fn` with node, `xhttp` from browser or our own custom *backend* to do so.

I guess is a matter of simplicity vs robustness. Just keep in mind that i18n has some features React Intl does not, and the "simplicity" I talk about is just in the configuration area. Both work in a very similar manner.

mxstbr:
> Yeah, it is detected by headers in request.

Oh, that's great! I think that's actually good enough, let's guess based on the users headers and then let them switch if it's the wrong one. I don't really see a point in linking somebody to a specific language, since they'll want to read it in their language anyway!

Another question that popped into my mind is what will happen with partial/unfinished translations? We wouldn't want to show those to users yet, but we do want to have them in the repo already.

victorhqc:
> Another question that popped into my mind is what will happen with partial/unfinished translations? We wouldn't want to show those to users yet, but we do want to have them in the repo already.

Good question 🤔 I'll check and post the result here.

victorhqc:
@philpl Another small difference is license. React intl has BSD 3 License and i18n has MIT not sure if that's important.

victorhqc:
@mxstbr Removed the `title` key from spanish translation and this happened

Missing translation

What do you think? Should we add a message to the user warning a missing translation somewhere? Or having the default `english` is ok?

mxstbr:
Oh, that's fine. We'll keep adding content to the website anyway, and I don't expect everything to just immediately be translated, so showing the english version is good enough. 

We should definitely have a CTA somewhere to improve the translations though!

Also, how does Google handle this i18n stuff? What do we have to consider there?

victorhqc:
My experience with SEO is limited. However, the translations are coming translated already from the server as this is page is rendered in the backend right? I guess we need to provide a way to google to crawl the different languages available. Further research needed there 🤔

victorhqc:
I found [this](https://support.google.com/webmasters/answer/182192?hl=en) about handling languages. Google recommend having multiple URLs for each language. So I think we'll need to add `nginx` rules to handle that too. On top of having just the headers for detecting language. The domain we'll be an important way to get the correct language too. And by domain I mean just the path. i.e. `styled-components.com/de`, `styled-components.com/en`, etc.

mxstbr:
Hmm, we deploy the website on now.sh so I don't think we have access to any nginx stuff? Can we do that from the backend?

victorhqc:
You're totally right. In that case we need to add the new routes in server. I'll update my PR to have that.

victorhqc:
@mxstbr Ok, this took more time than I thought. First of all, my previous code was wrong. I was using Javascript in the browser to get the translations. I deactivated `JS` in my browser to see if the translations were working and they don't. So I had to rewrite a bit what I did. Now `i18n` is starting in the server so it comes already translated. So you were right, some changes in `server.js` were needed. Now after that works, I'm back at creating additional paths for specific `languages`

victorhqc:
@philpl I found a good explanation with differences with `react-intl` [here](https://github.com/i18next/react-i18next/tree/master/example/nextjs)
> In contrast to react-intl the translations are visible both during development and in production

victorhqc:
So, current implementation has a way to set language by a parameter in URL. I guess we can make a dedicated `url` in `server.js` but need to read a bit more about that.

### English
English

### Spanish
screen shot 2017-10-26 at 2 35 40 pm

Disabled JavaScript, to make sure translations are happening in backend
screen shot 2017-10-26 at 2 37 19 pm

mxstbr:
That's awesome! We should be able to "alias" `/es/docs/basics` to `/docs/basics?lng=es` opaquely on the server so that the user (and Google) still see `/es/docs/basics` but the server just renders the `/docs/basics` router with that language?

victorhqc:
Exactly. That's what I have in mind :)

victorhqc:
I was a bit curious about how to translate comments in the website. Wasn't that hard. I'm worried about files with multiple lines though. Right now is just `json` files. It also supports `json5` and `yaml`. Here's how it looks the home page now. The comments are translated
Home

Also added tests for spanish language and it is rendered correctly in the `snapshots` 😄

mxstbr:
Yeah, let's try tackling some docs pages? I think those will be harder than the homepage

locales/es/homeGettingStarted.yml
```diff
@@ -0,0 +1,111 @@
...
+      `}
+    `;
+    ```
+
+    Todo lo que estamos diciendo aquí es que cuando la pripiedad `primary` esté puesta queremos que añada un poco más de `css` a nuestro componente,
+    en este caso, cambiar el color de fondo y el color.
+
+    Eso es todo, hemos acabado! Echa un vistazo al componente terminado
+
+    ```react
```
philpl: To put quick comment here, this is great to get the triple backticks without escaping, but

- the yaml wrapper is definitely not as nice, as I think it is with just having a tag in js
- this wouldn't allow us to use interpolations for JS 😢 

I'd personally say we should just make folders for each markdown with the appropriate files in there, translated and untouched apart from that

victorhqc: Yeah. This part was quite troublesome. Interpolation was totally dead and I had to extract the components and just render them. It gave good results though

victorhqc:
It's looking good!

Spanish homescreen

locales/es/homeGettingStarted.yml
victorhqc: Not sure how else can we get multiple line strings. Tried to use `json5` but failed and probably would have the same issues than `yaml`. Using just `json` was even worse, as all the lines had to be added in a single line with `\n`.

sections/homepage-getting-started.js
```diff
@@ -36,141 +37,75 @@ const SecondButton = styled.button`
...
+              <a href="https://spectrum.chat/styled-components">
+                <Badge
+                  src="https://withspectrum.github.io/badge/badge.svg"
+                  alt={translate('joinTheCommunity')}
+                />
+              </a>
+            </AlignCenter>
+
+            {md(translate('markdown.first'))}
+            <AlignCenter>
```
victorhqc: Here is where I had to render <AlignCenter>, <ExampleButton>, etc. directly in the component. As I couldn't use js interpolation. Which makes sense, we don't need that in the translation file.

victorhqc: Take a look @philpl 😄

philpl: I'm still wondering whether we should just split the file and make one per translation for content at least, like this and the docs pages :)

victorhqc: Good question. I'd suggest we split it per page, so we have a translation file per component page. As it's intuitive how to find the translations. The downside to that is that we might end up with a lot of translation files and keys and that could be hard to manage. So I'll leave that to decide @philpl and @mxstbr

victorhqc:
@mxstbr So I made some progress regarding the `url`. I managed to have the translations using the url path. So now we have `/es` and `/en` as index. as well as `/es/docs`, `/en/docs`

victorhqc:
Spanish is working with `/es/docs` because my browser has English by default

Spanish

If I just go to `/docs` I see it in english
English

I can go to `/en/docs` if I want
Forced English

mxstbr:
Should we just redirect `/en` to `/` to avoid Google thinking we're duplicating content?

victorhqc:
🤔 Makes sense. But just `/en` should be redirected right? `/es`, `/de`, etc. Should stay the same

victorhqc:
And how about `/en/docs` ? Should we redirect that to just `/docs` ?

mxstbr:
Yep, that's exactly it!

victorhqc:
Done, now every time there's a route containing `/en` at the beginning it will redirect it to the correct path.
`/en` -> `/`
`/en/docs` -> `/docs`
`/en/docs/basics` -> `/docs/basics`
`/en/docs/advanced` -> `/docs/advanced`
`/en/docs/api` -> `/docs/api`

If the route has another language, i.e. `es/docs` then that's left as it is.

mxstbr:
That is amazing! :clap:

sections/homepage-getting-started.js
victorhqc: @philpl So I went back to check the documentation. Unfortunately we're stuck using `namespaces` (`ns`), and the API only provides this as customization
```js
backend: {
  loadPath: __dirname + '/locales/{{lng}}/{{ns}}.yml'
}
```
So I think we can't have `locales/en/docs/first.yml` 😞

victorhqc: You can take a look [here](https://github.com/i18next/i18next-node-remote-backend)

victorhqc:
I'll proceed to extract all English (and translate to Spanish 😄) the entire site then. And add tests to missing pages.

victorhqc:
@mxstbr I extracted English from all the site and translated it to spanish. There are some minor issues bugging me though. Navigation in the `/docs` in one submenu (i.e. `/docs/basics#some-anchor`) doesn't work correctly. But if I'm already in `docs/basics` navigating with the navigation menu works perfectly. Not sure what's making the anchors to not work in this specific case. Even if I refresh the site the anchor seems to be working just fine.

mxstbr:
Mind updating the lock files so that builds pass?

screen shot 2017-11-01 at 3 22 40 pm

```sh
yarn
npm i
git add .
git commit -m 'Update lock files'
```

mxstbr:
Amazing work so far! 😍 

> Navigation in the /docs submenu (i.e. /docs/basics#some-anchor) doesn't work correctly.

/cc @philpl who built that

victorhqc:
Thanks 😄

philpl:
@victorhqc what's the problem with the anchors? If you point me at some links, I can take a look to see what's wrong.

Also, do we want to explore some other possibilities first that might not involve the yaml files? Otherwise this would be another mapping between the docs `md` pages and the yaml files that we have to maintain.

cc @mxstbr

victorhqc:
@philpl and about the `yaml` issue. I only encountered "problems" in the home page and API page, all the other `md` pages worked just fine because they're not using imported components.

```js
import SomeComponent from 'component`

md(`
 # Foo

 // This is impossible to map to an external string. Or not practical at least.
 ${<SomeComponent />}
`)
```

Almost all the files are just `md` without external components and work just fine

victorhqc:
@philpl we could extend the `md` component to support having external components being mapped to a string, but that's a separate issue all together. We would need to have some key tags in the translated string like `%$index%` so we could map the imported component to an index in the string somehow.

philpl:
@victorhqc 

> The impact of having a website in many languages is huge, specially for younger people that might not be super fluent in English like high school students.

I never doubted this in the first place :wink: Actually, the `md` tag refactor was meant to make this easier in the first place, and we just never got to actually start this.

> The yaml is not the problem, we could use json5 or any other format. There is no way to interpolate real components into an external string file. It works in master because the backtick is just sugar on top of the real implementation of how md is invoked.

I'm not talking about yaml / json5 or even adding components to the translations, but more of the other way around.

Since we have logic to route people to different files we can switch loading different files. So we can branch the `sections/` folder into different subfolders for languages and just create files like: `sections/en/faqs/nesting.js`.

This way we wouldn't need the translation files for the docs. It might not be optimal in some places, but I'm thinking more of maintainers of translations.

Translations notoriously go out-of-date quickly, and we  need to make the process simple. So if we have a page per translation:

- there can never be a situation of a changed page where the corresponding translation hasn't changed
- no mapping so the translator doesn't have to switch between three files: the page, the english translation, and their own language's translation
- trivial changes to allow different layouts for different languages, which might become relevant with more different languages that aren't European, like Japanese

For some languages, technical terms are either totally different, or have to stay in English and need some more references. So often things will need to diverge, and thus it's nice to keep everything in less files.

I hope this makes sense to you, and I'm not criticising the great work you've already done. I'm just aware that if we merge this now, and don't get it right, it'll only become harder to fix later on. :smile:

victorhqc:
@philpl That makes sense now. Thanks for that detailed response. So if I got your point correctly, you want to split the sections even further so its easier to maintain?

> Since we have logic to route people to different files we can switch loading different files. So we can branch the sections/ folder into different subfolders for languages and just create files like: sections/en/faqs/nesting.js.

I'm not sure if I follow you there. I saw that the sections have actually many subsections in there. You plan to split those sections into individual files?

And I still don't see how we can solve the issue about mapping external components into the `md` without doing the refactor.

victorhqc:
And about the `md` component. I'll think of something to fix this mapping issue. I'm thinking to pass the components currying the function, similar of what I did to send the `i18n` instance down to the `Link` component.

```js
const md = ({ i18n, components }) => props => (
 ...
)
```

philpl:
@victorhqc my idea is basically, instead of sourcing the translations for those sections from a separate file, can we decide at the page level which component to render.

In that case we'd probably have a component providing us with the current locale code, and can then import a different sections file depending on the translation.

Then we wouldn't have to insert the markdown code using the separate yaml files, which I think is preferable for the docs sections.

victorhqc:
So we'd be doing like a dynamic `import` depending on the current locale? Not sure if that'd work though. In server side it might work, but not sure in front end.

Correct me if I'm wrong but you suggest somehow to do something like this?

```js

const component = require(`{currentLanguage}/sections/faqs/nesting.js`)
```

Note that I used `require` and not `import` as the latter won't work with dynamic reference. And even with `require` I think is not encouraged. The current language is obtained on runtime, so having this in a bundle might be tricky.

philpl:
@victorhqc we can branch off a level upwards to avoid too many imports basically. But optimally we'd probably keep all the pages separate, and let next do the work:

- `pages/docs` becomes `pages/docs/en`
- Spanish docs pages will go into `pages/docs/es`
- Sections in `sections/x` become `sections/en/x` and so on

This would be quite static, but if it becomes too bad we can easily write a babel plugin to make this easier, which would work for both the client and server.

This should avoid dynamic imports, which is important since we don't want the content to be displayed asynchronously to the rest of the page in the docs.

Then the server redirects as it does now, essentially.

victorhqc:
@philpl  Interesting 🤔 that might actually work. If Next.js can handle it that is. I'll try this suggestion then. It might actually work 😄 
So just to make sure, is this accomplished using this?
https://github.com/zeit/next.js/#dynamic-import

I'm a newbie with next.js 😅

victorhqc:
@philpl I'll "revert" the extraction for the `md` files and have independent files for each language. If this works there might not be need to split the project further.

philpl:
@victorhqc no, I don't think we need to use dynamic imports, or even want to use them, since then the content would load separately when the client navigates to a new docs page. :/

victorhqc:
@philpl gotcha. Then how can we tell `next.js` what file to fetch? Is that declared in the `server.js` part?

```js
i18nextMiddleware.addRoute(i18n, '/:lng/docs/advanced', LANGUAGES, router, 'get', (req, res) => {
  if (isUsingDefaultLanguage(req)) {
    return redirectTo(res, `/docs/advanced/${DEFAULT_LANGUAGE}`)
  }

  return cachedRender(req, res, `/docs/advanced/${req.i18n.language}`)
})
```

philpl:
@victorhqc if we create the corresponding folders inside `pages/` then next will automatically recognise its route (not sure how much you know about this?)

so `/pages/docs/basics.js` could go into `/pages/docs/en/basics.js`, which then corresponds to the endpoint `/docs/en/basics` instead of `/docs/basics`. We can also split every page, i.e. `/pages/en/docs/basics.js` to make the route `/en/docs/basics` immediately.

Then we just have to set up redirects, i.e. `/docs/basics` 301 => `/LANG/docs/basics`

And for pages that are shared and have yaml translation files we could for example `pages/ecosystem.js` to be rendered maually, like you said, when the user accesses `/LANG/ecosystem`

victorhqc:
Interesting. All right, I'll try that out and update this thread when I have news. Thanks @philpl!

victorhqc:
@philpl Quick update. Seems like this approach kinda works. I can easily have translated docs pages now and I see the advantages you mentioned earlier. However I'm noticing some issues with this approach:
1. Duplicated components everywhere. Now instead of having a `locales/` paths where all the translations live. Now, most of the translations live in their own components, meaning I have duplicated code every time I'm using a `md()` component.
  - `docs/es/basics` - `docs/en/basics`
  - `sections/es/basics/...` - `sections/en/basics/...`
  - `sections/es/advanced/...` - `sections/en/advanced/...`
  Having future languages means this will be way more difficult to maintain.
2. I found this redirect quite hacky, I need to break how `next.js` works in order to manage the redirection from `docs/en/basics` to `en/docs/basics` as this later route doesn't exist. Maybe I have this last part wrong, but I don't see how it can be handled nicely.

So as a conclusion, maybe focusing on the `md()` refactor might be a better idea.

So, back to your previous bullet points:

- _there can never be a situation of a changed page where the corresponding translation hasn't changed_: This case is tricky to solve, even if we split the components per language, if there's a change in `english`, replications need to occur anyway with either approach.
- _no mapping so the translator doesn't have to switch between three files: the page, the english translation, and their own language's translation_: In the approach I propose, the translator doesn't need to know about the page implementation. Just focus on the translation itself as it's a manageable string. If we left the translations in the components themselves, then we'll never be able to truly extract the translations to a independent service as the components will be tightly coupled to the translation itself.
- _trivial changes to allow different layouts for different languages, which might become relevant with more different languages that aren't European, like Japanese_: This already happens with spanish, the grammar is very different from english and I have to put some components different than english, both approaches solve this issue.

philpl:
@victorhqc Hm, I see, maybe I'm just overthinking this, but let's run through the comments first.

> Having future languages means this will be way more difficult to maintain.

well, it's just files to be copied. The difference between adding more translation files, or copying the pages/sections is negligible imho.

> I found this redirect quite hacky, I need to break how next.js works in order to manage the redirection from docs/en/basics to en/docs/basics as this later route doesn't exist

I think we can change the folder structure to accommodate for this, as mentioned earlier. 🤔 

> This case is tricky to solve, even if we split the components per language, if there's a change in english, replications need to occur anyway with either approach.

It's a subtle difference. Replication can be done at any point after the change has been made. But when there's a mapping between component/page and the translation, then the change has to be immediate if it would otherwise break the mapping.

> In the approach I propose, the translator doesn't need to know about the page implementation. Just focus on the translation itself as it's a manageable string.

I think that's a valid point. But knowledge is always required when translating, and the sections are essentially just markdown too.

> If we left the translations in the components themselves, then we'll never be able to truly extract the translations to a independent service as the components will be tightly coupled to the translation itself.

There's a couple of points here, if I'm not mistaken.

The coupling here is on purpose as I've proposed, because that avoids the mapping, so it's purely the opposite of the current approach. This is fine for us because translations will occur on Github with contributors, not on an external service. This is because a lot of knowledge and help is required to translate technical docs.

> This already happens with spanish, the grammar is very different from english and I have to put some components different than english, both approaches solve this issue.

Let's take this as a quick example: https://github.com/victorhqc/styled-components-website/blob/2827a2ea62f9a52f2dedfd50139bdeccdf4a6855/sections/basics/motivation.js#L48-L51

This example is a bit crude and not a hard one to solve, but let's suppose that it makes sense to add more markdown after the video for one translation. Then we'd have to introduce a new key.

What if the new key is not handled in other translations? What if we need to make some styling changes?

A lot of questions branch out there for me.

Originally we wanted to look at how `md` can be replaced with pure markdown files, and that's still a possibility. The only question to answer is, what happens with interpolations?

We can maybe solve it by providing a mapping

```js
{
  Video: ARandomComponent
}
```

to markdown and then write a `HtmlBlock` renderer

@mxstbr: I'm clearly overthinking a lot of bits here. What do you think? Is it worth going with translation files with a key-based mapping for now? Migrate later to pure markdown? Or use separate sections files? What sounds most appealing to you?

victorhqc:
@philpl thanks for your response. I see your point clearer now. I guess both are very different approaches, but both accomplish the same result in a different way. Both have their pro and cons. About the breaking mapping I'm not sure if I follow.

In the example I showed before in this thread, I removed a key from `spanish` and it got replaced by the `english` counterpart. For sure we'd need to be careful when adding/removing english keys as that change will affect all the languages.

The approach of having separate files per language will have a different problem though, if the English document is updated, the different languages will have an obsolete documentation, so I guess we have to evaluate what's worse:
1. Have an untranslated, but updated documentation text.
2. Have a complete translated page, but obsolete.

victorhqc:
@philpl 
> I think we can change the folder structure to accommodate for this, as mentioned earlier. 🤔

Changing the structure makes sense, we'd avoid the hacky redirection.

> This example is a bit crude and not a hard one to solve, but let's suppose that it makes sense to add more markdown after the video for one translation. Then we'd have to introduce a new key.
What if the new key is not handled in other translations? What if we need to make some styling changes?

That is a valid point too 🤔 So no perfect solution either way. Though I don't see why would some language would need to add text where other doesn't. This of course happens when we have text and links, a paragraph might change completely or the link has to be in a different position due to grammar.

But when we have a `Table`, a `Video` or whatever component that is not part of the text itself (like links) this never happens.  There are a lot of clever ways to avoid doing that using grammar and structuring the text to make it logical to avoid having text after the video.

victorhqc:
@mxstbr Feedback here would be nice 😄 I see two possible solutions, with two different set of pros and cons. If you can add something else, I'd be happy to hear it.

victorhqc:
Closing this. Too bad it wasn't what you were looking for

philpl:
@victorhqc hey, sorry about that 😅 let’s see what we can do to make our desired approach more defined. When the time comes I’m sure we can pick up parts of this PR, if not even the entire thing 🙂

Issue #116, opened 2017-09-26
Theming section is missing information on the "manual" theme prop


philpl:
It's possible to circumvent a missing `ThemeProvider` or override it with the `theme` prop.

We're not noting that the `theme` prop is reserved and how it can be used.

https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L110

tejasbubane:
I think this can be closed now.

mxstbr:
Yep, good catch, thanks @tejasbubane!

PR #149, opened 2017-10-25
Docs for the `theme` prop

ucorina:
Here's my attempt at https://github.com/styled-components/styled-components-website/issues/116.

I thought a bit about whether to add this info as part of the main theming section (the first one) or as a separate one and went for a separate one in the end, to keep the explanations from growing too much.
Did you have anything else in mind on where to put this?

Also, I didn't actually mention anything about the prop being reserved. Should I?

PS. You'll notice I changed the `line-height` of the code blocks - that's to keep the text in `monospace` font aligned with everything else in titles and content text. 

Looking forward to feedback & further instructions :) Thanks!

### DIFF
--- a/components/Code.js
+++ b/components/Code.js
@@ -10,6 +10,7 @@ const Code = styled.span`
   font-weight: 300;
   padding: 0 0.05em 0.1em;
   vertical-align: bottom;
+  line-height: 1.4;
 `
 
 export default Code
--- a/sections/advanced/theming.js
+++ b/sections/advanced/theming.js
@@ -41,7 +41,7 @@ const Theming = () => md`
 
       <ThemeProvider theme={theme}>
         <Button>Themed</Button>
-      </ThemeProvider>
+      </ThemeProvider>      
     </div>
   );
   \`\`\`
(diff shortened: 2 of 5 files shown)


ucorina:
Whoa, this was fast! Thanks a lot for the review!

mxstbr:
Thanks a lot for contributing, I didn't really do anything! :blush:

components/Code.js
```diff
@@ -10,6 +10,7 @@ const Code = styled.span`
   font-weight: 300;
   padding: 0 0.05em 0.1em;
   vertical-align: bottom;
+  line-height: 1.4;
```
philpl: @ucorina @mxstbr Hiya, not sure if this was intentional? I think the alignment of the inline-code element took a while, so I'm not sure whether this will turn out correct, or whether this changes it at all

mxstbr: > that's to keep the text in monospace font aligned with everything else in titles and content text.

philpl: Cheers, must’ve overlooked that 😅 apologies

PR #151, opened 2017-10-26
Added Coinbase logo

jorilallo:
screen shot 2017-10-26 at 11 51 06 am

We're big fans of styled components at Coinbase. We been using them for a while now for Coinbase.com.

Source: Frontend engineer there


morajabi:
@jorilallo You should tell us when to buy which ICO then :)

PR #154, opened 2017-10-31
Improve WithIsScrolled's responsiveness

philpl:
- Remove debounce (both in component and the package)
- Prevent setState when the isScrolled state won't change
- Replace PureComponent as it's not necessary anymore
- Replace render check with onMount invariant
- Pass this.state instead of controlled props object to children()


morajabi:
btw what do you think about making the site static? @philpl @mxstbr

mxstbr:
We very soon will have that logic with #148 I think?

philpl:
@mxstbr @morajabi yea, we need to keep the server and dynamic build around for the translations and redirections. But since we're caching SSR results and are on Zeit Now, there's not really any need to build the site statically.

What we can do though is prebuild the SSR cache and don't expire/rerun any renders during the server's runtime.

mxstbr:
> What we can do though is prebuild the SSR cache and don't expire/rerun any renders during the server's runtime.

We cache in Cloudflare for 24h, so `/ecosystem` refreshes once every 24 hours. It'd be great to have that auto-refresh? Otherwise it only updates when we re-deploy which doesn't seem to be happening very often...

philpl:
@mxstbr true; all in all I think the current caching setup with cloud flare + SSR LRU works well

morajabi:
@philpl @mxstbr Great, fair enough! Just I'm too biased :) haha

PR #161, opened 2017-11-02
Fix broken anchor links in SideBarMenu

chucksellick:
The sidebar links to https://www.styled-components.com/docs/advanced#server-side-rendering but the actual anchor in the doc itself was being rendered as https://www.styled-components.com/docs/advanced#server-side-rendering- (an extra dash at the end).

This is because the SSR heading has the ` | v2` tag appended, the string was being split on the pipe but the extra spaces weren't being trimmed.

### DIFF
--- a/utils/titleToDash.js
+++ b/utils/titleToDash.js
@@ -2,6 +2,7 @@ import elementToText from './elementToText'
 
 const titleToDash = title => (
   elementToText(title)
+    .trim()
     .toLowerCase()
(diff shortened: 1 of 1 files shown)


morajabi:
This stringifier does not want to be fixed hahaha 😆

chucksellick:
Created a new pull request with change as requested:
https://github.com/styled-components/styled-components-website/pull/164

mxstbr:
Thank you @chucksellick!

PR #166, opened 2017-11-06
Updated FAQ with question about DOM having two classes

### LINKED ISSUE #147: Add "Why do my DOM nodes have two classes?" to the FAQ
This is copied from a recent chat log, we should clean this up and add it to the FAQ:

> The first class (`sc-bfhe32`) is static per-component, meaning every instance of a `<Button />` will have that class. The second is per-dynamic-stylr-fragmet, so all `<Button primary />` will have the same one, all `<Button secodary />` will have the same one etc.
> Note that the first one doesn't actually have any styles attached to it because it's only used for when you interpolate the component (`${Button}`) and so that in the DevTools you can quickly tweak all `Button`s no matter if they're primary, secondary or whatever else

nilsgoksor:
Fix #147

Created a new file dom-two-classes.js in faqs answering the question "Why do my DOM nodes have two classes?".

Updated docs.json and faqs.js accordingly.


sections/faqs/dom-two-classes.js
```diff
@@ -0,0 +1,13 @@
+import md from 'components/md'
+
+const TwoDomClasses = () => md`
+  ## Why does a DOM node have two classes?
```
philpl: Can we change this to:

> Why do my DOM nodes have two classes?

otherwise it sounds like all DOM nodes are being referred to, regardless of whether they're mounted by styled components or not

sections/faqs/dom-two-classes.js
```diff
@@ -0,0 +1,13 @@
...
+
+  Each node actually has two classes connected to it;
+
+  One is static per component meaning that each instance of a component has this class, but it hasn't any style attached to it. However, it can be used to help the developer smoothly tweak all instances of a class at once.
```
philpl: > per *styled* component

> component*,* meaning

> each ~instance~ element of ~a~ *the styled component*.

> . But it *doesn't have* any style*s* attached to it.

philpl: > However, it can be used to help the developer smoothly tweak all instances of a class at once

This is not the actual intent. It is used to quickly **identify** the component, and to be able to make some quick changes in the DevTools.

sections/faqs/dom-two-classes.js
```diff
@@ -0,0 +1,13 @@
...
+
+  The other is dynamic, and it's this class which contains all the styling for specific components of a class. This is the class you usually modify.
```
philpl: > ~and it's this class which contains all the styling for specific components of a class~

let's say something along the lines of

> and this one will be different depending on what the interpolations result in. They will be different for every element of your styled component with different props.

Also:

> ~This is the class you usually modify.~

This would imply that it would be directly modified.

sections/faqs/dom-two-classes.js
```diff
@@ -0,0 +1,13 @@
...
+
+`
```
philpl: Let's also take some of @mxstbr's suggestions into account:

> The first class (sc-bfhe32) is static per-component, meaning every instance of a <Button /> will have that class. The second is per-dynamic-stylr-fragmet, so all <Button primary /> will have the same one, all <Button secodary /> will have the same one etc.
Note that the first one doesn't actually have any styles attached to it because it's only used for when you interpolate the component (${Button}) and so that in the DevTools you can quickly tweak all Buttons no matter if they're primary, secondary or whatever else

- It'd be great to show `<Button />` and `<Button secondary />` as examples of how the dynamic class can change
- It'd also be cool to show some examples of how the static and dynamic class can look

Apart from that we can also link to this section: https://www.styled-components.com/docs/tooling#better-debugging

Which helps the user to make the static class human readable for development

sections/faqs/dom-two-classes.js
nilsgoksor: Sure, I agree it's better.

sections/faqs/dom-two-classes.js
nilsgoksor: Alright, I guess I misunderstood a little bit. But speaking as a "noob", I would not understand what's meant by "it identifies the component quickly". 

Would you mind elaborating because I still don't understand/know what to write here.

sections/faqs/dom-two-classes.js
nilsgoksor: Based on your comment in the thread about the issue I thought you did not want an example. I agree it would be nice to show examples, but you don't want them to be live, or what did you mean by your comment? :)

sections/faqs/dom-two-classes.js
philpl: @nilsgoksor identify, as in, looking at the DOM element and knowing due to the class which ones come from the same styled component

sections/faqs/dom-two-classes.js
philpl: @nilsgoksor I mean examples as in inline code snippets, not as full fledged live editors or code blocks.

It could be included as:

> Given a styled component, for instance `<Button />`, it would render with the same static class every time. But if the styles are changed using interpolations, with something like `<Button secondary />`, then the dynamic class will be a different one, while the static class would remain the same.

morajabi: @nilsgoksor @philpl I agree inline code blocks are enough. Although nobody could gain anything from a live example :)

nilsgoksor:
Thanks for the feedback. I updated the description accordingly, let me know what you think :)

mxstbr:
What the hell is wrong with GitHub, I just resolved the conflicts and now it's got new conflicts? WTF

mxstbr:
Thank you @nilsgoksor, this is awesome!

PR #168, opened 2017-11-10
Add screenshot integration tests and remove page snapshots

### LINKED ISSUE #167: Remove docs snapshots
I'm not sure when we've introduced the snapshots for docs pages, but that doesn't add any value for testing, since it mostly just pollutes the PR with information that we don't need.

It'd be useful to replace this with an automated test that checks all docs pages in one checking whether:

- they render without an error
- they don't render the placeholder: https://github.com/styled-components/styled-components-website/blob/master/components/md.js#L25

Apart from that it'd be neat to use an image snapshotter for the docs page, which should actually save us some time: https://github.com/americanexpress/jest-image-snapshot

philpl:
Fix #167


philpl:
@MicheleBertoli just letting you know that this removes the section/page snapshots, since they're quite bloaty in PRs :smile:

morajabi:
@philpl I really like removing those snapshots! Awesome! One thing, can you explain briefly why we need screenshot integration tests? Thanks!

morajabi:
Why do we need to have a dot before jest.config.js?

philpl:
@morajabi yep, and I do hate dotfiles as well. We can move all configs to a folder soon. The reason for making it a dotfile is that it’s then hidden so I don’t get annoyed by them 😂

morajabi:
@philpl That was the reply of the day hahah 😂

mxstbr:
LGTM, just gotta fix CI hah

philpl:
@mxstbr @morajabi It passed and is ready for rereview :tada: :heart: some details:

- The fonts are reset to Arial and Courier (for code) to ensure consistency between build envrionments
- All animations are disabled because.........TIMING
- Chrome is installed on Travis (often it's already installed which saves time)
- When installing puppeteer a postbuild task will run on your devices that downloads a consitent version of chrome (can be skipped with an env variable)
- The screenshots are rendered at 600x400 at 1x resolution, which is the smallest we can do until they start failing on different OS'
- The threshold is 5% of pixels at most can be different
- It runs in Travis' non-docker env which gives a little more RAM and perf
- The tests are `integration` tests and separated

The last point means that you'll now need to run `yarn integration -u` to update the screenshots. This is not done automatically as it's slower than unit tests should be.

Travis now typically takes **3-4 min** which is AWESOME considering that it runs all tests, linters, and checks the screenshots.

mxstbr:
One thing that's up for discussion I guess: Do we want to screenshot all docs pages? By screenshotting them any change in content will need a snapshot update, which is kinda annoying, especially for first contributors.

What we want to know with the screenshots is: Did somebody break a layout? That means we could basically only screenshot compare two pages: The homepage, and one docs page. Then for the rest of the docs pages we only need to make sure that they render without an error, which doesn't need a screenshot but await page.waitForSelector('something-rendered-from-react').

WDYT?

philpl:
@mxstbr that's an excellent idea! Maybe we should be looking at the basics page? That one should contain all kinds of components

mxstbr:
Sure, sounds good to me!

philpl:
@mxstbr This now only screenshots the basics page (first docs page) and for all of the docs pages it searches for a marker and checks whether it exists. It also now collects runtime errors and fails when uncaught runtime errors occur.

components/DocsLayout.js
```diff
@@ -61,6 +61,9 @@ class DocsLayout extends Component {
 
           {children}
         </Content>
+
+        {/* used for integration tests */}
+        <div className="end-of-docs" />
```
mxstbr: I'd add a `data-e2e-id` to the root `Content` node instead, otherwise people might remove this eventually and `data-e2e-id` is more explicit :blush:

philpl:
@mxstbr done! changed the integration selectors to attribute selectors, and managed to add the homepage by overriding the hero's styling :+1:

integration-test/pages.spec.js
```diff
@@ -0,0 +1,110 @@
...
+    await app.start(PORT, 'localhost')
+    browser = await launch({
+      executablePath: process.env.GOOGLE_CHROME_BINARY || undefined,
+      args: ['--no-sandbox', '--disable-setuid-sandbox', '--disable-lcd-text']
+    })
+    page = await browser.newPage()
+
+    page.on('request', req => {
+      if (!req.url.startsWith(`http://localhost:${PORT}`)) {
+        console.log(req.url)
```
mxstbr: console.log?

philpl: ah darn it; I'll remove it once I squash and rebase 😆

philpl:
@mxstbr regarding the images; no extra resources will load including extra css files, scripts, images, etc... This is to prevent them from influencing the snapshots.

mxstbr:
Oh that makes sense, alright.

philpl:
Done :heart: this was half fun half madness 😆

mxstbr:
Awesome!!

PR #173, opened 2017-11-11
feat: add image proxy for img shield badges 

### LINKED ISSUE #171: Proxy shields.io badge
The shields.io badge takes a looooong time to load. We should proxy it through styled-components.com/downloads.svg or something so that we can cache the file in CloudFlare for 24h.

morajabi:
Fixes #171

I used `axios`

### DIFF
--- a/sections/homepage-getting-started.js
+++ b/sections/homepage-getting-started.js
@@ -39,13 +39,13 @@ const HomepageGettingStarted = () => (
   <Content>
     <AlignCenter>
       <a href="https://github.com/styled-components/styled-components">
-        <Badge src="https://img.shields.io/github/stars/styled-components/styled-components.svg?style=social&label=Star&maxAge=3600" alt="Stars on GitHub" />
+        <Badge src="/r/stars.svg" alt="Stars on GitHub" />
       </a>
       <a href="https://www.npmjs.com/package/styled-components">
-        <Badge src="https://img.shields.io/npm/v/styled-components.svg" alt="Current version" />
+        <Badge src="/r/npm-v.svg" alt="Current version" />
       </a>
-      <Badge src="https://img.shields.io/npm/dm/styled-components.svg?maxAge=3600" alt="Monthly downloads" />
-      <Badge src="http://img.badgesize.io/https://unpkg.com/styled-components/dist/styled-components.min.js?compression=gzip&label=gzip%20size" alt="Gzipped size" />
+      <Badge src="/r/downloads.svg" alt="Monthly downloads" />
+      <Badge src="/r/size.svg" alt="Gzipped size" />
       <a href="https://spectrum.chat/styled-components">
         <Badge alt="Join the community on Spectrum" src="https://withspectrum.github.io/badge/badge.svg" />
       </a>
--- a/sections/homepage-getting-started.js
+++ b/sections/homepage-getting-started.js
@@ -39,13 +39,13 @@ const HomepageGettingStarted = () => (
   <Content>
     <AlignCenter>
       <a href="https://github.com/styled-components/styled-components">
-        <Badge src="/r/stars.svg" alt="Stars on GitHub" />
+        <Badge src="/proxy/stars.svg" alt="Stars on GitHub" />
       </a>
       <a href="https://www.npmjs.com/package/styled-components">
-        <Badge src="/r/npm-v.svg" alt="Current version" />
+        <Badge src="/proxy/npm-v.svg" alt="Current version" />
       </a>
-      <Badge src="/r/downloads.svg" alt="Monthly downloads" />
-      <Badge src="/r/size.svg" alt="Gzipped size" />
+      <Badge src="/proxy/downloads.svg" alt="Monthly downloads" />
+      <Badge src="/proxy/size.svg" alt="Gzipped size" />
       <a href="https://spectrum.chat/styled-components">
         <Badge alt="Join the community on Spectrum" src="https://withspectrum.github.io/badge/badge.svg" />
       </a>
--- a/server.js
+++ b/server.js
@@ -66,7 +66,7 @@ const cachedProxyServer = (req, res, imgUrl, remoteUrl) => {
     res.end(data, 'binary')
   }).catch(() => {
     // Failed to download image
-    next()
+    res.status(500).send('Error')
   })
 }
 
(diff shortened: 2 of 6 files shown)


morajabi:
OK! just wanted to be clear it's not a normal image.

server.js
```diff
@@ -65,8 +65,41 @@ app.prepare()
       cachedRender(req, res, '/docs/api')
     })
 
-    server.get('/sw.js', (req, res) => {
```
philpl: This was accidentally deleted I suppose :)

server.js
```diff
@@ -65,8 +65,41 @@ app.prepare()
...
+
+      // Check if we want to proxy this
+      if (typeof remoteUrl === 'undefined') {
+        // Let NextJS handle it (edither a route or 404 error)
+        // (P.S It's funny we call `next()` to pass it to Next 😆)
+        next()
+      }
+
+      try {
+        const { data } = await axios.get(remoteUrl, {
```
philpl: This is blocking the request until the server is done fetching the image, but we could also just stream it.

Also, this does the request over and over again instead of utilising an LRU cache, like the SSR stuff :)

server.js
```diff
@@ -65,8 +65,41 @@ app.prepare()
...
+          responseType: 'arraybuffer'
+        })
+        res.type('image/svg+xml')
```
philpl: This only applies to the resources we have now, but it might be problematic in the future.

Maybe this should just use one of the many express proxy packages that apply all this logic already? 🤔

philpl:
@morajabi I don’t think it’s quite ready yet :)

Regarding Max’s comment, I do think as well that /r is a rather random name (haha, get it?! *r*andom... I’ll see myself out)

We could just do /proxy really.

server.js
morajabi: Suggested one over Twitter, probably you missed it, but yea I wanted to keep it as simple as possible.

server.js
morajabi: Ohhhhh, sorted

morajabi:
@philpl You've got really good eyes 👏 , I need to practice on that.

philpl:
👀 let's get this merged

morajabi:
👀

sections/homepage-getting-started.js
```diff
@@ -39,13 +39,13 @@ const HomepageGettingStarted = () => (
...
-        <Badge src="https://img.shields.io/github/stars/styled-components/styled-components.svg?style=social&label=Star&maxAge=3600" alt="Stars on GitHub" />
+        <Badge src="/r/stars.svg" alt="Stars on GitHub" />
       </a>
       <a href="https://www.npmjs.com/package/styled-components">
-        <Badge src="https://img.shields.io/npm/v/styled-components.svg" alt="Current version" />
+        <Badge src="/r/npm-v.svg" alt="Current version" />
       </a>
-      <Badge src="https://img.shields.io/npm/dm/styled-components.svg?maxAge=3600" alt="Monthly downloads" />
-      <Badge src="http://img.badgesize.io/https://unpkg.com/styled-components/dist/styled-components.min.js?compression=gzip&label=gzip%20size" alt="Gzipped size" />
+      <Badge src="/r/downloads.svg" alt="Monthly downloads" />
```
philpl: The URLs here haven't been updated

morajabi: AhhhhhhhhhhhhhhhhhhhhhhhhhhhhX100000

server.js
```diff
@@ -1,6 +1,5 @@
 const dev = process.env.NODE_ENV !== 'production'
 const moduleAlias = require('module-alias')
-const path = require('path')
```
philpl: This will need to be readded for `/sw.js`! :wink:

philpl:
@morajabi So a couple of things:

- Found a couple of bugs (see review)
- Instead of a 404 a 500 is thrown for images which don't exist
- Several comments should really be deleted i.e. (`// Are we sure about `?maxAge=3600`?`, and `// (P.S It's funny we call `next()` to pass it to Next 😆 )` :wink:)
- `res.append('X-Cache', 'MISS')` is missing meaning we don't see a `HIT`/`MISS` header

morajabi:
what do you mean by res.append('X-Cache', 'MISS') is missing meaning we don't see a HIT/MISS header

philpl:
@morajabi so for the `cachedRender` call we append a new header to the response, which is called `X-Cache`, a really common name for cache headers. By setting it to `HIT` (LRU was used) or `MISS` (LRU was not used & updated), we can easily see whether we're getting a cached response or an actual one.

morajabi:
@philpl I'm using that just like you did.

morajabi:
"Instead of a 404 a 500 is thrown for images which don't exist" Wrong. I pass non-existents to `next`

morajabi:
👀

philpl:
@morajabi Haha I don't mean to frustrate you, sorry. But you'll have to hold on a little longer.

Regarding the forwarding, I guess the object existence check is insufficient:

```
$ http get 'http://localhost:3000/proxy/xyz.svg'

HTTP/1.1 500 Internal Server Error
Connection: keep-alive
Content-Length: 5
Content-Type: text/html; charset=utf-8
Date: Mon, 13 Nov 2017 15:16:02 GMT
ETag: W/"5-fy9qFc+NorJ+Wkr0e1jnrXHAs9k"
X-Powered-By: Next.js 2.4.8
Error
```

As you can see this goes through your code and not the next.js handler. But you'll just need to send a 404. There's no need to use the request like a middleware, it'll just need to send a 404.

Regarding the caching; the header is there indeed, but it seems that I'm never receiving a cached response:

```
$ http get 'http://localhost:3000/proxy/size.svg'

HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 961
Content-Type: image/svg+xml;charset=utf-8
Date: Mon, 13 Nov 2017 15:18:45 GMT
X-Cache: MISS
```

morajabi:
@philpl are you sure you are running the server not by `yarn dev`?

philpl:
@morajabi oh, you're right. I done goofed. So the caching works.

So only the 404 code is left.

morajabi:
done.

morajabi:
@philpl  DO NOT ENCOURAGE BAD PRACTICES 😄  I don't have eslint installed globally.

morajabi:
What do you mean we missed it in lint-staged?

morajabi:
@philpl I'm going to buy some cookies 🍪  for you and let's celebrate this huge success!

PR #176, opened 2017-11-21
Added note on nested components and the distinction between RC and SC

oliverturner:
Addresses #1314

### DIFF
--- a/sections/faqs/reverse-selectors.js
+++ b/sections/faqs/reverse-selectors.js
@@ -1,4 +1,4 @@
-import md from "components/md";
+import md from "components/md"
 
 const ReverseSelectors = () => md`
   ## Can I refer to other components?
@@ -58,21 +58,21 @@ const ReverseSelectors = () => md`
   attempting to mount \`B\` in the following example will fail because component 
   \`A\` is an instance of React.Component not a Styled Component.
 
-  ```react
+  \`\`\`jsx
   class A extends React.Component {
     render() {
       return <div />;
     }
   }
   
-  const B = styled.div`
-    ${A} {
+  const B = styled.div\`
+    \${A} {
     }
-  `;
-  ```
+  \`;
+  \`\`\`
 
   The error thrown - \`Cannot call a class as a function\` - occurs because the 
   styled component is attempting to call the component as an interpolation function.
-`;
+`
 
-export default ReverseSelectors;
+export default ReverseSelectors
--- a/sections/faqs/reverse-selectors.js
+++ b/sections/faqs/reverse-selectors.js
@@ -54,9 +54,10 @@ const ReverseSelectors = () => md`
 
   ### Caveat
 
-  This behaviour is only supported within the context of *Styled* Components: attempting to mount 
-  `B` in the following example will fail (`Cannot call a class as a function`) because the component 
-  we're trying to extend is an instance of React.Component not a Styled Component
+  This behaviour is only supported within the context of *Styled* Components: 
+  attempting to mount `B` in the following example will fail (`Cannot call a 
+  class as a function`) because component `A` is an instance of React.Component 
+  not a Styled Component.
 
   \`\`\`react
   class A extends React.Component<{}> {
--- a/sections/faqs/reverse-selectors.js
+++ b/sections/faqs/reverse-selectors.js
@@ -55,8 +55,8 @@ const ReverseSelectors = () => md`
   ### Caveat
   
   This behaviour is only supported within the context of *Styled* Components: 
-  attempting to mount `B` in the following example will fail because component 
-  `A` is an instance of React.Component not a Styled Component.
+  attempting to mount \`B\` in the following example will fail because component 
+  \`A\` is an instance of React.Component not a Styled Component.
 
   ```react
   class A extends React.Component {
@@ -71,7 +71,7 @@ const ReverseSelectors = () => md`
   `;
   ```
 
-  The error thrown - `Cannot call a class as a function` - occurs because the 
+  The error thrown - \`Cannot call a class as a function\` - occurs because the 
   styled component is attempting to call the component as an interpolation function.
 `;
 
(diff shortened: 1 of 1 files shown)


sections/faqs/reverse-selectors.js
```diff
@@ -51,6 +51,25 @@ const ReverseSelectors = () => md`
...
-`
 
-export default ReverseSelectors
+  ### Caveat
+
+  This behaviour is only supported within the context of *Styled* Components: attempting to mount 
+  `B` in the following example will fail (`Cannot call a class as a function`) because the referenced
+  component is an instance of React.Component not a Styled Component
+
+  \`\`\`react
```
philpl: We probably don't want this to be a live-editable example, since it won't work :smile:

sections/faqs/reverse-selectors.js
```diff
@@ -51,6 +51,25 @@ const ReverseSelectors = () => md`
 
   We could have nested the color-changing rule within our Link component, but then we'd have to
   consider both sets of rules to understand why Icon behaves as it does.
-`
 
-export default ReverseSelectors
+  ### Caveat
```
philpl: I'm not sure whether we'll need to add another heading here 🤔

sections/faqs/reverse-selectors.js
```diff
@@ -51,6 +51,25 @@ const ReverseSelectors = () => md`
...
+
+  This behaviour is only supported within the context of *Styled* Components: attempting to mount 
+  `B` in the following example will fail (`Cannot call a class as a function`) because the referenced
+  component is an instance of React.Component not a Styled Component
```
philpl: Missing full stop :wink:

sections/faqs/reverse-selectors.js
```diff
@@ -51,6 +51,25 @@ const ReverseSelectors = () => md`
...
+
+  \`\`\`react
+  class A extends React.Component<{}> {
```
philpl: The flow/TS annotation can be removed

sections/faqs/reverse-selectors.js
oliverturner: It's a good heading Phil.

More seriously, where we're receiving bug reports about misconceptions around this behaviour I'm inclined to be emphatic.

sections/faqs/reverse-selectors.js
oliverturner: Ugh. Good catch.

sections/faqs/reverse-selectors.js
oliverturner: D'oh 👍

sections/faqs/reverse-selectors.js
philpl: Alright :+1:

sections/faqs/reverse-selectors.js
oliverturner: TIL this is the fence for a live example!

sections/faqs/reverse-selectors.js
```diff
@@ -51,6 +51,25 @@ const ReverseSelectors = () => md`
 
   We could have nested the color-changing rule within our Link component, but then we'd have to
   consider both sets of rules to understand why Icon behaves as it does.
-`
 
-export default ReverseSelectors
+  ### Caveat
+
+  This behaviour is only supported within the context of *Styled* Components: attempting to mount 
+  `B` in the following example will fail (`Cannot call a class as a function`) because the referenced
```
philpl: > (`Cannot call a class as a function`)

Maybe we can add another sentence saying something along the lines of:

> It will also throw an error saying "cannot call a class as a function," because the styled component will attempt to call the component as an interpolation function.

sections/faqs/reverse-selectors.js
```diff
@@ -53,14 +53,13 @@ const ReverseSelectors = () => md`
   consider both sets of rules to understand why Icon behaves as it does.
 
   ### Caveat
-
+  
   This behaviour is only supported within the context of *Styled* Components: 
-  attempting to mount `B` in the following example will fail (`Cannot call a 
-  class as a function`) because component `A` is an instance of React.Component 
-  not a Styled Component.
+  attempting to mount `B` in the following example will fail because component
```
philpl: Is the colon really necessary?

> ...supported within the context of *Styled* Components. Attempting to mount

sections/faqs/reverse-selectors.js
```diff
@@ -53,14 +53,13 @@ const ReverseSelectors = () => md`
...
   This behaviour is only supported within the context of *Styled* Components: 
-  attempting to mount `B` in the following example will fail (`Cannot call a 
-  class as a function`) because component `A` is an instance of React.Component 
-  not a Styled Component.
+  attempting to mount `B` in the following example will fail because component 
+  `A` is an instance of React.Component not a Styled Component.
 
-  \`\`\`react
-  class A extends React.Component<{}> {
+  ```react
```
philpl: The escapes are missing and we probably want to use:

```
\`\`\`jsx
```

sections/faqs/reverse-selectors.js
```diff
@@ -70,7 +69,10 @@ const ReverseSelectors = () => md`
     ${A} {
     }
   `;
-  \`\`\`
+  ```
+
+  The error thrown - `Cannot call a class as a function` - occurs because the
```
philpl: Small punctuation nit; Can we use an em dash here? "—"

philpl: @oliverturner this nit is still outstanding; looks fine to me otherwise :+1:

oliverturner: Seems not: commonmark doesn't support the `---` syntax for emdash and my editor treats -  and — as the same character :(

philpl: wut; ok, that's weird

philpl:
@oliverturner Thanks again :heart: great work!

PR #178, opened 2017-11-30
Documentation for appending styles to target

bteng22:
@mxstbr Opening PR to add documentation for `Targeted Styles` for https://github.com/styled-components/styled-components/pull/1324

Hopefully this PR can start a conversation regarding how we want to expose this feature. Definitely open to feedback or questions!

### DIFF
--- a/pages/docs.json
+++ b/pages/docs.json
@@ -60,6 +60,9 @@
         },
         {
           "title": "Server Side Rendering"
+        },
+        {
+          "title": "Targeted Styles"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


bteng22:
`Optionally you can use the StyleSheetManager provider directly, instead of this method. **Just make sure not to use it on the client-side.**` This is noted on the [Server Side Rendering](https://www.styled-components.com/docs/advanced#server-side-rendering) section. 

Is this no longer the case with #1324 since we're using `StyleSheetManager` on the client-side?

PR #179, opened 2017-11-30
Fix attrs example

glennreyes:
I think this should be `padding`


mxstbr:
Something's wrong with this CI, it shouldn't be failing. /cc @philpl

PR #193, opened 2017-12-19
Try screenshot testing (Do not merge!)

philpl:


mxstbr:
If this is indeed the issue than we really need to make that clearer.

Maybe we should integrate Danger into the repo and have it post the diff if the tests fail due to the screenshots?

mxstbr:
I mean, I'm not going to say I told you that's not the issue, but I told you that's not the issue. 😜😜😜

philpl:
@mxstbr I think it'd be great to add danger and output a hint at running `yarn integration -u` :+1:

Well, it's not like I read through everything before testing this 😆

PR #199, opened 2017-12-27
Higher specificity faq

GGAlanSmithee:
Closes #187 cc @mxstbr I simply took your explanation from https://github.com/styled-components/styled-components/issues/1253#issuecomment-337871693 and adjusted it a bit.

PR #200, opened 2017-12-27
Override inline styles

GGAlanSmithee:
This builds on #199 and adds another section about overriding inline styles.

In the section I describe using the CSS style `element-attr` with `!important` since it seems to be the recommended way to override inline styles. In all honesty, I am having a hard time seeing how it differs from simply using `!important`. It might be more backwards compatible, but I cannot verify this.

So, we can either leave this as is, since it's sure to work, or we can simplify it to only use `!important`.

Sources
* https://css-tricks.com/override-inline-styles-with-css/
* https://stackoverflow.com/questions/16813220/how-can-i-override-inline-styles-with-external-css
* https://www.sitepoint.com/override-inline-css/

### DIFF
--- a/pages/docs.json
+++ b/pages/docs.json
@@ -116,6 +116,7 @@
         { "title": "Can I refer to other components?" },
         { "title": "When should I use styled()?" },
         { "title": "Can I use css frameworks?" },
+        { "title": "How can I override styles with higher specificity?" },
         { "title": "Why do my DOM nodes have two classes?" }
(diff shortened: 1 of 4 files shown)


GGAlanSmithee:
For changes only related to inline-css, see commit https://github.com/styled-components/styled-components-website/pull/200/commits/66d01c7629859d0c9e9ebb6569c9335008768cc1

PR #207, opened 2018-01-09
Add a section to FAQs on when to use attrs

kadikraman:
As suggested on [Twitter](https://twitter.com/mxstbr/status/950289161227030528)

### DIFF
--- a/sections/faqs/when-to-use-attrs.js
+++ b/sections/faqs/when-to-use-attrs.js
@@ -19,6 +19,10 @@ const WhenToUseAttrs = () => md`
   // This specific one is hidden, so let's set aria-hidden
   <PasswordInput aria-hidden="true" />
   \`\`\`
+
+  The same goes for props that can be inferred based on the "mode" of another prop.
+  In this case you can set a property on \`attrs\` to a function that computes
+  that prop based on other props.
 `
 
 export default WhenToUseAttrs
--- a/sections/faqs/when-to-use-attrs.js
+++ b/sections/faqs/when-to-use-attrs.js
@@ -3,7 +3,7 @@ import md from 'components/md'
 const WhenToUseAttrs = () => md`
   ## When to use attrs?
 
-  You can pass in attributes to styled components using \`attrs\`, but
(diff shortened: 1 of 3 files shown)


sections/faqs/when-to-use-attrs.js
```diff
@@ -0,0 +1,24 @@
+import md from 'components/md'
+
+const WhenToUseAttrs = () => md`
+  ## When to use attrs?
+
+  You can pass in attributes to styled components using \`attrs\`, but
```
philpl: I'd add a link to the corresponding API page here somewhere :smile:

sections/faqs/when-to-use-attrs.js
```diff
@@ -0,0 +1,24 @@
...
+  different one:
+
+  \`\`\`js
+  const PasswordInput = styled.input.attrs({
+    // Every <PasswordInput /> should be type="password"
+    type: 'password'
+  })\`\`
+
+  // This specific one is hidden, so let's set aria-hidden
+  <PasswordInput aria-hidden="true" />
```
philpl: Maybe we can also add a paragraph on the lines of:

> The same goes for props that will be inferred based on the “mode” or another prop that are being passed. In this case you can set a property on `attrs` to a function that computes that prop based on other props.

kadikraman: Ah, so you mean if you have a prop that's calculated based on another prop, you can do it in `atts` like this?
```js
const MyComponent = styled(SomeComponent).attrs({
  param: props => props.someOtherParam ? 1 : 0
)}``;
```

I can't think of a nice simple real life example of this at the moment. Can you? 🤔

philpl: Yep, it'd be nice if it's mentioned.

I think a real life example would be `<Button disabled onClick={x} />` since one could override `onClick` when disabled is set to true. But I do not think this needs an example :wink:

kadikraman: Fair enough. I've added a paragraph on this as requested, thanks Phil ✨

mxstbr:

PR #211, opened 2018-01-10
update Release page copy

### LINKED ISSUE #204: Missing dot in the first paragraph on the Releases page (+ writing style question)
> Updating styled components is usually just an npm install, only major version might have breaking changes that will be noted in the CHANGELOG

We should add a full stop at the end of the sentence. 

Also, what do you think about making the `CHANGELOG` capital case or lowercase?

halvves:
fixes #204


SaraVieira:
The not approving was a mistake, fat fingers sorry

PR #217, opened 2018-01-14
Accessibility issues

ink8bit:
# A11y issues
Hi, 
I found some accessibility issues and decided to make a PR.

## Changes
-  Added `lang` attribute to `html` tag (see more [info](https://dequeuniversity.com/rules/axe/2.2/html-has-lang))
- Fixed an issue with some links which didn't have discernible names (see more [info](https://dequeuniversity.com/rules/axe/2.2/link-name))

> I'm not so sure about my realisation of the second issue (**w3c** offers this [solution](https://www.w3.org/TR/WCAG20-TECHS/C7.html)). Maybe you can help me find a better solution.

Hope this would be helpful.

### DIFF
--- a/components/Anchor.js
+++ b/components/Anchor.js
@@ -25,6 +25,14 @@ const Anchor = styled.a`
   position: absolute;
   left: 0;
   color: inherit;
+
+  & span {
+    height: 1px;
+    width: 1px;
+    position: absolute;
+    overflow: hidden;
+    top: -10px;
+  }
 `
 
 const AnchorIcon = styled(LinkIcon).attrs({
@@ -70,6 +78,7 @@ const Link = ({ children, id, sub }) => {
 
       <Anchor href={`#${id}`}>
         <AnchorIcon />
+        <span>Anchor link:{id}</span>
       </Anchor>
 
       {children}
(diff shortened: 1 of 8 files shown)


components/Anchor.js
```diff
@@ -70,6 +78,7 @@ const Link = ({ children, id, sub }) => {
 
       <Anchor href={`#${id}`}>
         <AnchorIcon />
+        <span>Anchor link:{id}</span>
```
philpl: I'm not sure whether this is actually the best way to solve this? Wouldn't it be possible to set an `aria` attribute on the anchor itself?

It seems unclean to render a `width/height: 1px` span on each of the anchors

ink8bit: I'm afraid that this is the only one way to do it

probablyup: The `title` attribute should be sufficient for all anchor tags.

probablyup:
@ink8bit would love to get this in if you have time to address the PR comments?

ink8bit:
@probablyup Hi! Sorry for the delay.
Yeah, would be great if you can do it. Have no time to deal with it.

PR #219, opened 2018-01-15
Update docs for server side rendering with example for Apollo GraphQL

niklaskorz:
Add description and example for using styled-components ssr with react-apollo ssr

### DIFF
--- a/sections/advanced/server-side-rendering.js
+++ b/sections/advanced/server-side-rendering.js
@@ -55,6 +55,35 @@ const ServerSideRendering = () => md`
   You'll also need to customize the \`.babelrc\` and use \`babel-plugin-styled-components\`.
 
   Refer to [our example](https://github.com/zeit/next.js/tree/master/examples/with-styled-components) in the Next.js repo for an up-to-date usage example.
+
+  ### Apollo GraphQL
+
+  When using server side rendering with both styled-components and
+  [Apollo Client](https://www.apollographql.com/docs/react/recipes/server-side-rendering.html),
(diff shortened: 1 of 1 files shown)


sections/advanced/server-side-rendering.js
```diff
@@ -55,6 +55,35 @@ const ServerSideRendering = () => md`
   You'll also need to customize the \`.babelrc\` and use \`babel-plugin-styled-components\`.
 
   Refer to [our example](https://github.com/zeit/next.js/tree/master/examples/with-styled-components) in the Next.js repo for an up-to-date usage example.
+
+  ### Apollo GraphQL
+
+  When using server side rendering with both styled-components and
+  [Apollo Client](https://www.apollographql.com/docs/react/recipes/server-side-rendering.html),
+  make sure that not only calls to React's render method are wrapped by a \`StyleSheetManager\`,
+  but calls to [getDataFromTree](https://www.apollographql.com/docs/react/recipes/server-side-rendering.html#getDataFromTree)
```
kitten: I think this might not 100% belong in our docs, but since it's a short section, I don't see a reason to reject it :)

That being said:

> make sure that not only calls to React's render method are wrapped by a StyleSheetManager,
> but calls to getDataFromTree as well

Sorry, I'm not quite sure I know what you'd like to express here. I've implemented the combination of the two before, but this isn't quite matching up with what we see.

Would you rather say:

> the `ServerStyleSheet` instance needs to be evaluated after `getDataFromTree` has run; This means that you need to move your `getStyleTags` after `renderToString` and not only after (and definitely not before) `getDataFromTree`.

I think that'd be more descriptive. But the example does the job nicely anyway :)

PR #224, opened 2018-01-23
Update react-native docs for v3.0.0

pyros2097:
We need to import
`import styled from "styled-components";`
instead of
`import styled from "styled-components/native";`


philpl:
Sorry, but we already have a PR for it ready ☹️ https://github.com/styled-components/styled-components-website/pull/223

PR #226, opened 2018-01-25
Clarify that SSR methods only work after wrapped components are rendered

not-an-aardvark:
This clarifies that it's not possible to use tags from `sheet.getStyleTags()`  in the same tree as the component that `sheet` is wrapping. For example, one could mistakenly try to do something like this to include their app and the generated `<style>` tags as part of a larger component, but in fact this does not work:

```js
const wrapperConstructor = () => {
  const stylesheet = new ServerStyleSheet();
  const wrappedComponent = stylesheet.collectStyles(<YourApp />);

  return (
    <div>
      {stylesheet.getStyleElement()}
      {wrappedComponent}
    </div>
  );
};
```

### DIFF
--- a/sections/advanced/server-side-rendering.js
+++ b/sections/advanced/server-side-rendering.js
@@ -46,6 +46,8 @@ const ServerSideRendering = () => md`
   Alternatively the \`ServerStyleSheet\` instance also has a \`getStyleElement()\` method
   that returns an array of React elements.
 
+  Note that \`sheet.getStyleTags()\` and \`sheet.getStyleElement()\` can only be called after your element is rendered. As a result, components from \`sheet.getStyleElement()\` cannot be combined with \`<YourApp />\` into a larger component.
+
   ### Next.js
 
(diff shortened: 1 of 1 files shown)


sections/advanced/server-side-rendering.js
```diff
@@ -46,6 +46,8 @@ const ServerSideRendering = () => md`
   Alternatively the \`ServerStyleSheet\` instance also has a \`getStyleElement()\` method
   that returns an array of React elements.
 
+  Note that \`sheet.getStyleTags()\` and \`sheet.getStyleElement()\` can only be called after your element is rendered. As a result, components from \`sheet.getStyleElement()\` cannot be combined with \`<YourApp />\` into a larger component.
```
philpl: Can you prefix this with ">" and remove "Note ..."

This will render it as an actual note :wink::+1:

philpl:
Cheers :+1:

mxstbr:

PR #227, opened 2018-01-26
Update StyleSheetManager component usage example

ericberens:


sections/advanced/server-side-rendering.js
```diff
@@ -32,7 +32,7 @@ const ServerSideRendering = () => md`
 
   const sheet = new ServerStyleSheet()
   const html = renderToString(
-    <StyleSheetManager sheet={sheet.instance}>
```
ericberens: This is actually correct but the typings are incorrect.

PR #230, opened 2018-02-04
Add OpenCollective sponsors to homepage

osamajandali:
- Create new react component for sponsors section.
- Use axios to make the request.
- Filter tier by sponsors only.
- If the sponsor doesn’t have logo there is a svg logo rendered instead.


morajabi:
Can you deploy it to now.sh?

sections/homepage-sponsors.js
```diff
@@ -0,0 +1,77 @@
...
+function checkImageLink(imageLink){
+  return /^(http|https)/.test(imageLink)
+}
+
+class HomepageSponsors extends Component {
+  state = {
+    sponsors: []
+  }
+
+  componentDidMount() {
```
kitten: Can you copy the logic that we have for our releases page please? There’s some tricks in Next that will SSR the content which is definitely beneficial in this case

Also can you not install axioms while that page is using fetch please?

osamajandali: I have done it like this but I have run into an issue : 
- *getInitialProps* is only invoked on pages so do I have to add it on index or just use it via *componentDidMount* instead.
- *axios* is already there used on server and I haven't install it so I do not think that I have to uninstall it.

kitten: @osamajandali 

> getInitialProps is only invoked on pages so do I have to add it on index or just get the fetched data via componentDidMount instead.

Yea, we would have to add it to index and pass it down, which I think is fair.

> axios is already there used on server and I haven't install it so I do not think that I have to uninstall it.

On the server, yes, but not on the client. It'd be great to match the github fetcher to not increase the bundle size unnecessarily :wink: https://github.com/styled-components/styled-components-website/blob/master/utils/githubApi.js

The Releases page code is also what the hook for index.js would look like: https://github.com/styled-components/styled-components-website/blob/master/pages/releases.js#L46-L52

kitten:
Btw I'm not sure whether it's a good choice to add the list to the end of the homepage. Due to it being totally unrelated to the "Getting Started" guide above and the length of that guide we might as well add it to a separate, dedicated page. cc @mxstbr

sections/homepage-sponsors.js
osamajandali: Yeah I have already done with this thanks for your help. I will wait to see our final decision whether we keep it on index or move it before I update this PR.

mxstbr:
I think we should still highlight them on the homepage, but I feel like it's weird to have both the "users" list and the "sponsor" list next to each other with the Sponsor list being less important?

mxstbr:
We could totally also add a separate page for this though!

kitten:
@morajabi I think this would play right into your skills :wink: do you have an idea of how we can highlight our OpenCollective sponsors best without just lovelessly adding a section to the homepage?

morajabi:
Let's talk tonight.

osamajandali:
Guys, I am doing some mockups for the sponsors section now and I am wondering are we able to have some tweaks on getting started section **(if needed)** in order for it to consistent with the sponsor section ?

mxstbr:
Yeah sure, depending on the tweaks necessary

osamajandali:
I have been doing some stuff for awhile now and here are my ideas about this:
- Move **Getting Started** to a separate page and adding the sponsors directly after the hero.
- Break **Getting Started** into sections and put **Sponsors** within these sections ( for example we add **installation** section then sponsors then some more sections)
- Add **Sponsors** section at the end of the getting started but not with different style so we could tell that they are not related.
- Separate page for **Sponsors**.
I am really into second option and currently working on a design for it, I just want to hear your opinion guys.

morajabi:
Just as other friends said, it's a bit awkward we emphasise companies that just use styled-components, more than sponsors. I think this should be figured out in a clever way. 

We used Babel docs as an inspiration before, now as an example, let's see how [Babel](http://babeljs.io/) respects sponsors. They have that giant section for sponsors *before* "Who's using Babel?" section which is at the end.

@osamajandali's second suggestion is close to what babel did. We have badges + install section (with "Install" heading) then sponsors, and after that, "Getting Started" section comes. We *can* keep the "USED BY FOLKS AT" section there as it is too if @mxstbr or @kitten prefer.

osamajandali:
@morajabi That's cool that we agree we can work on this together if you'd like.

morajabi:
@osamajandali That's open source 😉 

If you can come up with some rough design, it'd be really cool since I'm a bit busy currently. If not, we still appreciate your work and we'll do this next week.

kitten:
@mxstbr @morajabi 

> it's a bit awkward we emphasise companies that just use styled-components, more than sponsors

I think this is a good point, but right now, as we're not emphasising Open Collective as much, we just don't have a large number of sponsors yet. But it will be important to emphasise the difference and to highlight sponsors in a positive light.

@osamajandali

> Move Getting Started to a separate page and adding the sponsors directly after the hero.

I actually really like this idea, but I think @mxstbr had the point of it being useful for getting started *right away*. As in, entering the site, reading on, not having to click anything, and knowing how it works. I'm not sure whether this is the right thing to do anymore. It'd be trivial to improve this section on a separate page and show some snippets first instead, i.e. slimming down this section and moving the actual "intro guide" somewhere else.

mxstbr:
Okay, let's remove users from homepage and replace them with sponsors, (since why would anybody sponsor otherwise?) and then as a second step add a `/users` page which only has logos but doesn't link to anything. (where sponsors, on the other hand, can link to whatever page they want)

I really don't want to remove getting started from the homepage fwiw

morajabi:
@mxstbr @kitten I think you a bit misunderstood what was said regarding the getting started. As I didn't say to remove/move it.

But we don't need to remove users either, just push them down to the bottom of the page just like Babel website, and make them black `filter: brightness(0) invert(0)`. 

The only reason I'd say to keep users there, is if we want to attract new adopters. Then the approach can be as I said before (we don't remove Getting Started):

>@osamajandali's second suggestion is close to what babel did. We have badges + install section (with "Install" heading) then sponsors, and after that, "Getting Started" section comes. We can keep the "USED BY FOLKS AT" section there as it is too if @mxstbr or @kitten prefer.

osamajandali:
I think we need a final decision here @mxstbr @kitten , now we have many more options :
- Replace the users with sponsor (We need here a white background as you may have logos with a white background)
- Reorganize getting started as me and @morajabi suggesting

mxstbr:
I like @morajabi's suggestion of moving users to the bottom and adding a section add the top with the sponsors. Let's do that!

osamajandali:
I have made a mockup of some tweaks that I feel it's necessary for the page design balance after adding sponsors section, take a look:
styled component

kitten:
@osamajandali that looks awesome! I’d be happy to go ahead with this design. I’ll leave the decision up to @morajabi since he’s designed the last overhaul and hence has the best idea of what works well with the rest of the site 👍

mxstbr:
Let's leave the headings left aligned though, there's no need to change the entire typography of the website. Other than that, love it, let's build it!

morajabi:
@mxstbr I agree, we should keep headings consistent across the site.
@osamajandali Thanks for this, awesome! I have a few little things in mind, overall it's very good:
- Let's make remove the dark grey background for "Used by ..." section and add a line above it, to separate it
- Did you forget the badges? 😄

osamajandali:
@morajabi Thanks for your feedback
But I think the users section have a special style that needs a special design. Also, I already tried white background and I didn't like it at all, take a look:
styled component

So then I thought that dark grey color is good as it matches with the editor color. So how this sounds to you ?

morajabi:
Lower the logos contrast to better match the footer background, add a border around the box and center it. Add a padding to the box so logos don’t touch the edge. Keep in mind users section is not much important thus we don’t need to grab much attention :) If it’s not clear, I’ll prepare a quick sketch for you.

osamajandali:
@morajabi @mxstbr do you like what I did of swapping the colors of link (pink) and header (orange) and decreasing the font size a bit ?

osamajandali:
@morajabi yeah give me a sketch in order to get what you mean

osamajandali:
@morajabi Is this what you mean ?
styled component

morajabi:
@osamajandali Yeah I had exactly this in mind! Just a little radius to match the site and lower the border contrast a bit more. We can try removing the footer background later as well. Go for it! It would be trivial to add tweaks or followup PRs in any case :)

osamajandali:
@mxstbr @kitten we need review here guys if you have time 😁

osamajandali:
@mxstbr I copied the section as it is without playing with the padding only some padding between logos and the borders, shall I decrease any of them? I feel that it looks good!

sections/homepage-sponsors.js
```diff
@@ -0,0 +1,56 @@
...
+)
+
+function checkImageLink(imageLink){
+  return /^(http|https)/.test(imageLink)
+}
+
+const HomepageSponsors = ({ sponsors }) => (
+  <Wrapper>
+    {sponsors.map(sponsor => 
+      sponsor.tier == "Sponsor" ?
```
kitten: I'm not sure whether this loading state is quite correct; is this meant to filter instead?

btw I've tried starting this up and it's not quite loading for me :(

> invalid json response body at https://opencollective.com/styled-components/members/all.json reason: Unexpected token < in JSON at position 0

Maybe you're missing a header for the request? Or am I missing something?

Also, the loading indicators seem to be showing for each sponsor (maybe not a good idea) and those indicators are `position: absolute` right now unfortunately, so they're all over the place

osamajandali: I have fixed this loading state.
I tried to have to replicate this error  **Unexpected token < in JSON at position 0** but i haven't got anything, everything is working just fine.
Try to remove this check **sponsor.tier == "Sponsor" ?** check you should see all the contributors by then.

osamajandali: @kitten take a look here 😄

probablyup:
@osamajandali if you could get this rebased, I'd be happy to land it

mxstbr:
I'd love to only enable it for silver and gold sponsors ($500 / $2000+ respectively) in order to make it more enticing to support us financially ("Get your logo with a link on the s-c homepage!!!").

Also, I think we should land #275 in conjunction with this, as that would remove the need for any more logos on the homepage. Just add a showcase module to the homepage which points out some of the pages using styled-components and then link to the showcase page.

el1f:
Hey guys! What's blocking this from being implemented? Any way I can help? :)

PR #238, opened 2018-02-28
Add note on using stylelint v9+

emilgoldsmith:
I will be merging in https://github.com/styled-components/stylelint-processor-styled-components/pull/167 any second now and publishing it, and the addition should be self explanatory as to why it is added


kitten:
Let's wait until it's actually published, just in case?

emilgoldsmith:
published it! had a commitment i had to go to a few minutes before the CI finished

PR #247, opened 2018-03-16
Update stylelint.js

corysimmons:
Make destructive `--fix` warning more prominent/descriptive.
Also, fml.

### DIFF
--- a/sections/tooling/stylelint.js
+++ b/sections/tooling/stylelint.js
@@ -1,7 +1,7 @@
 import md from 'components/md'
 
 const StylelintProcessor = () => md`
-> Beware that due to limitations on what is possible for Stylelint custom processors we cannot support the \`--fix\` option. ⚠️ **If you try to use \`--fix\` you will destroy most of your unsaved work.**
+> Beware that due to limitations on what is possible for Stylelint custom processors we cannot support the \`--fix\` option. ⚠️ **If you try to use \`--fix\` you might destroy most of your unsaved work.**
 
 ## Stylelint
 
--- a/sections/tooling/stylelint.js
+++ b/sections/tooling/stylelint.js
@@ -1,6 +1,8 @@
 import md from 'components/md'
 
 const StylelintProcessor = () => md`
+> Beware that due to limitations on what is possible for Stylelint custom processors we cannot support the \`--fix\` option. ⚠️ **If you try to use \`--fix\` you will destroy most of your unsaved work.**
+
 ## Stylelint
 
   Lint your [styled components](https://github.com/styled-components/styled-components) with [stylelint](http://stylelint.io/)!
@@ -55,8 +57,6 @@ Now you can lint your CSS by running the script! 🎉
 \`\`\`
 npm run lint:css
 \`\`\`
-
-> Beware that due to limitations on what is possible for Stylelint custom processors we cannot support the \`--fix\` option
 
 #### Webpack
 


sections/tooling/stylelint.js
```diff
@@ -1,6 +1,8 @@
 import md from 'components/md'
 
 const StylelintProcessor = () => md`
+> Beware that due to limitations on what is possible for Stylelint custom processors we cannot support the \`--fix\` option. ⚠️ **If you try to use \`--fix\` you will destroy most of your unsaved work.**
```
mxstbr: Let's change `will` to `might`.

Also, super sorry this happened to you, that's a shitty experience for sure. 😕 Maybe there's something we can do from the processor side to detect whether you've used `--fix` and block stylelint from running if that's the case? /cc @emilgoldsmith

corysimmons:
@mxstbr updated.

It's okay, my fault for not saving often enough, but yeah, thought stylelint would just hit the css in my `.js` and instead it literally just wiped entire files. lol

emilgoldsmith:
Also, while it is of course important to make the dangers clear, and I'm sorry "for your loss", what about rewording it to the exact effect instead of the "drama"? As in:

> Using `--fix` will overwrite your js files with the css contained in those files due to the limitations of the Stylelint processor API

Or something like that, I think the "destroying most of your unsaved work" is unnecessarily dramatic (though again, I'm sorry for all the work you lost).

I'm not sure if we from the processor API can detect if people have supplied the `--fix` option? Because in that case we could maybe both issue a warning and then just output the js file so as to avoid these unfortunate incidents. As far as I know it is not provided though, right? It'd be a pretty easy PR to the Stylelint repo to add that argument though I could imagine, if anyone wants to take it up

emilgoldsmith:
If you want to add in `which could result in you losing unsaved work` to what I said above I'm also okay with that, I just feel the current wording is vague + dramatic as it doesn't actually say what's actually happening

emilgoldsmith:
Sorry for the spam, but also just thought, an even easier PR into Stylelint would to be to make a quick change that checks if there are any processors in your config, and if there are + `--fix` was supplied then an error is shown and nothing is run as, as far as I can think of, there isn't a single processor that could make `--fix` work since it would literally only be on plain CSS and then no processor would be needed. This also moves the responsibility to Stylelint and will benefit all processors until Stylelint provides a `--fix` API, instead of having every processor developer having to implement something like that

corysimmons:
This PR is a good example of maintainers driving people away from contributing.

The point is, when people use stylelint, they are excited to use `--fix` to autoformat stuff (prettier has spoiled us).

You have docs that allude to people using stylelint, so it's a good bet they're going to try to use `--fix` which will straight up WIPE entire projects.

Please make a more prominent warning somehow/someway.

emilgoldsmith:
I'm sorry you feel that way @corysimmons, I even actually just submitted [this issue](https://github.com/stylelint/stylelint/issues/3223) while you were writing that. 

I really wasn't trying to drive you away from contributing, just trying to keep the documentation as helpful as possible, and was really trying to think of ways to solve your problem for the future. I definitely think moving up the warning to the top is useful, and if you don't feel like reopening your PR I might make one myself, if you want to help I think (given that the Stylelint maintainers agree with implementing it) a great way would be to fix the whole problem in the issue I opened above.

kitten:
@emilgoldsmith yes, I’d love to include a warning like this in the docs while adding an error is on the way. I’m personally not using stylelint but this sounds like a dangerous gotcha.

@corysimmons Apologies from my end for the misunderstanding, but PRs are inherently places for discussions when they haven’t been preceded by issues. This means they’ll be under scrutiny and I hope that it won’t keep you from contributing in the future. However, I’d much rather prefer a PR being improved and questioned without it being merged without any discussion happening at all

emilgoldsmith:
@corysimmons the issue on Stylelint got accepted by the way, should be a pretty quick PR if you still want to help out, otherwise we can hope someone picks it up soon :)

PR #248, opened 2018-03-21
Add import line to show users where to import `ThemeProvider` from

vviikk:
There are a few `ThemeProvider` components on the web that are misleading. Including `react-theme-provider` which has similar syntax. Would be good to see this mentioned so users know it's built into styled-components.


sections/advanced/theming.js
```diff
@@ -117,6 +121,8 @@ const Theming = () => md`
   This is useful to circumvent a missing \`ThemeProvider\` or to override it.
 
   \`\`\`react
+  import styled, { ThemeProvider } from 'styled-components';
```
kitten: This won't work as the `react` markdown snippets show a live editor, which doesn't support import statements right now.

I'd be happy to add a static `jsx` snippet with just the line of the import above the first example.

But I'm not sure as to whether this import isn't quite obvious :wink: Anyway, it'd be great to just add a single `jsx` snippet for this

piggyslasher: Hi, it might seem really obvious if you've used styled components before. But the ThemeProvider is a name that isn't unique to this project. Would difinitely save a few poor devs precious minutes.

I've updated the PR. How's that?

sections/advanced/theming.js
```diff
@@ -10,6 +10,12 @@ const Theming = () => md`
   To illustrate this, let's create our Button component, but this time we'll pass some variables down
   as a theme.
 
+  Firstly, you would need to import the \`theme-provider\` wrapper component.
```
probablyup: nit: First~~ly~~, you ~~would~~ need

piggyslasher:
Bump. Sorry. Let's get this merged?

PR #253, opened 2018-04-05
Advanced media template has extra `/16`

terrierscript:
I try to use advanced's media template but not work well.
I found the strange calculate `${sizes[label] / 16}`.

What is `/16` ? I guess it's maybe typo.


terrierscript:
sorry, It's my misreading

PR #257, opened 2018-05-20
Document using placeholder <style> tag

TrevorBurnham:
As discussed at https://github.com/styled-components/styled-components/issues/1707. This technique is relevant to the existing "Issues with Specificity" section, as an alternative to increasing the specificity of selectors in stylesheets.


sections/advanced/existing-css.js
```diff
@@ -72,6 +70,18 @@ const ExistingCSS = () => md`
...
+  competing tags. That means either moving your stylesheet \`<link>\` tags down into the \`<body>\`,
+  or moving the styled-components \`<style>\` tag(s) up. You can control where styled-components
+  renders its styles by creating a placeholder tag with the \`data-styled-components\` attribute:
+
+  \`\`\`html
+  <head>
+    <style data-styled-components="true">/* runtime styles will go here */</style>
+    <link rel="stylesheet" href="my-component.css"/>
+  </head>
+  \`\`\`
```
kitten: Let's remove the comment here. It doesn't really add any value and it's not quite factually correct, as styled-components might append more style tags after it :wink:

kitten: Let's also add a note that this works because of SSR rehydration. Something like:

> This tag is also used in server-side rendering. During runtime styled-components searches for this tag and recycles it, adding new style tags in its place.

TrevorBurnham: > Let's remove the comment here

Done: ff6af8a

> Let's also add a note that this works because of SSR rehydration

Sure, how's this? a0e3f7d

mxstbr:

sections/advanced/existing-css.js
brendonco: Is this going to be in v5++?

PR #258, opened 2018-05-21
Document techniques for avoiding conflicts when embedded

TrevorBurnham:
This PR adds a new section under "Issues with Specificity" about potential problems when embedding styled-components in a page you don't control, and the recommended solutions:

1. To give styled component rules higher specificity than the host page's style rules, use [babel-plugin-styled-components-css-namespace](https://github.com/QuickBase/babel-plugin-styled-components-css-namespace)
1. To avoid potential conflicts with any other instance of styled-components that might be on the page, override `SC_ATTR` (using the functionality recently added by https://github.com/styled-components/styled-components/pull/1632)


sections/advanced/existing-css.js
```diff
@@ -86,6 +86,41 @@ const ExistingCSS = () => md`
   This technique takes advantage of styled-components’ server-side rendering (SSR) support. At
   runtime, styled-components searches the document for existing tags and recycles them. Any new
   \`<style>\` tags are inserted immediately after the existing ones.
+
+  ### Avoiding Conflicts When Embedded
```
probablyup: Perhaps this could be reworded to "Avoiding Conflicts With Existing Page Styles"?

TrevorBurnham: Hmm. I'm not sure that would clearly delineate this section from the preceding section, "Issues with Specificity." This section is specifically about issues that arise when styled-components is on a page that you don't control, so that e.g. there could be style rules you don't know about, and there could be another instance of styled-components.

I realize that the term "embedding" is overloaded with meanings, so I'd welcome other suggestions.

probablyup: "Using Styled Components on Third Party Sites" possibly?

TrevorBurnham: Not sure that's more accurate. It could be your own site but with third-party CSS or JavaScript, for instance. Of course, "embedded" isn't accurate in that scenario, either.

How about "Avoiding Conflicts with Third-Party Styles and Scripts"?

probablyup:
Sure, yeah that makes sense.

TrevorBurnham:
@probablyup OK, I put in that heading change, and I've rebased this branch against master to keep the history tidy.

PR #263, opened 2018-05-31
Add a note to remind usage of JS-style comments

chinesedfan:
People are easy to be confused about the usage of JS-style comments. Like the discussion in https://github.com/styled-components/stylelint-processor-styled-components/issues/195.

@jeddy3 @emilgoldsmith Please correct me if I made a mistake. (cc @mvasin @Undistraction)


Undistraction:
I would say something like:

> Although JS-style comments (`// example`) are supported by styled-components, they are invalid in CSS. This means that by default, stylelint will flag them as errors. To work around this problem, the syntax can be changed to 'scss' which does support js-style comments. This can only be done using  a flag passed in to the stylelint cli: 
> ```bash
> stylelint './src/**/*.js' --syntax scss
> ```
> Important: You cannot change the syntax in a `.stylelintrc` file.
> 

I also wonder if doing this has any side-effects. Will it mean that scss syntax that wouldn't be valid in styled-components will now be able to slip through?

emilgoldsmith:
What are the cases where this has been a problem exactly though? Because I remember someone bringing this up earlier, and I wrote a unit test and checked everything and couldn't reproduce, and we aren't using anything in config or CLI options etc.

mvasin:
What I as a user still have hard time to understand, if thanks to `stylelint-processor-styled-components` stylelint starts to tolerate weird (from CSS point of view) constructs like `${props => ...}`, why `//` comments can't be packed into the same pack without extra step on our side?

There are no variations on SC syntax, so why do we need tweak syntax-related options?

Undistraction:
@emilgoldsmith Problem is that JS comments don't work out the box and flag errors despite being valid in styled-component CSS.

The solution is very much nonobvious so needs to documented, however a better solution would be for stylelint-processor-styled-components to add the support in the same way it adds support for all the other esoteric syntax used by styled components.

emilgoldsmith:
I definitely agree to good documentation and making READMEs as helpful as possible, my point was just that as far as we can see we DO support it, and I would like to know more about in what way it is breaking. @mvasin provided some helpful information in the issue now though, which clears things up a bit, and at least clarifies a reproduceable way to find the problem (though I haven't tried yet, but I trust it should be reproduceable)

chinesedfan:
@emilgoldsmith I have already created a PR, https://github.com/styled-components/stylelint-processor-styled-components/pull/197, to explain the reason why tests passed.

PR #269, opened 2018-06-11
Feature/recommended reading

oliverturner:
As promised, a PR that adds a reference to @taliahm's amazing review of SC


oliverturner:
Updated navigation: clicking "Coming from CSS" wasn't working because the heading "How do Styled Components work within a component?" didn't match.

oliverturner:
Aww... this is the nicest org to makes PRs to!

PR #281, opened 2018-07-15
Accessibility

jmiazga:
I used Lighthouse to identify and address some accessibility issues. It does not address the following issues:

- Background and foreground colors do not have a sufficient contrast ratio.
- [id] attributes on the page are not unique

I plan on opening an Issue around ```[id] attributes on the page are not unique```

### DIFF
--- a/components/Anchor.js
+++ b/components/Anchor.js
@@ -80,7 +80,7 @@ const Link = ({ children, level, id }) => {
     <Child>
       <InvisibleAnchor id={id} />
 
-      <Anchor href={`#${id}`}>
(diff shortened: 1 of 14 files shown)


kitten:
@probablyup it seems that the deploy task on Travis is triggering on non-master branches and is obviously failing in PRs 🤔

@jmiazga awesome! I think this is good to go. If you continue to work on the other issues let’s make sure to run some polls on Twitter or sth so that people actually like the new colours 👏 so hit me up if you end up doing that

jmiazga:
@kitten I plan on combing through it with a few more tools. I'm guessing there will be a couple more small PRs in the future.

PR #252, opened 2018-04-01
Update typescript.js

christianalfoni:
I went through a long process trying to figure this out. I hope this documentation will help others to just use TypeScript out of the box :)

PR #283, opened 2018-07-26
fix escaping on Typescript docs

christianalfoni:
Okay, I believe this should work now :)

https://github.com/styled-components/styled-components-website/pull/252#issuecomment-405678860


christianalfoni:
Okay, updated the branch as well... hopefully all okay now :)

PR #288, opened 2018-08-15
emphasize babel plugin usage in the installation & getting started sections

probablyup:
Also removed a reference to bower since it's defunct.


sections/homepage-getting-started.js
```diff
@@ -58,18 +61,33 @@ const HomepageGettingStarted = () => (
       To download styled-components run \`npm install --save styled-components\`.
       That's all you need to do, you are now ready to use it in your app! (yep, no build step needed 👌)
 
+      > It's highly recommended to also use the [styled-components babel plugin](https://github.com/styled-components/babel-plugin-styled-components). It offers many benefits like more legible class names, server-side rendering compatibility, smaller bundles, and more.
```
mxstbr: Let's make sure to make it clear that it's not recommended, maybe something like

> If you do have access to the build configuration and are using Babel it's recommended to use the styled-components babel plugin. (it is not required to use styled-components though, unless you're server-side rendering) It offers many benefits like more legible class names, server-side rendering compatibility, smaller bundles, and more.

probablyup:
No it IS recommended. We WANT people to use the plugin. It gives too many
benefits not to.

mxstbr:
A I brain farted, I meant is not **required** of course. Edited the comment, wdyt of my wording?

probablyup:
@mxstbr done

PR #289, opened 2018-08-21
revise docs away from StyledComponent.extend to styled(StyledComponent)

probablyup:


probablyup:
@mxstbr made a bunch of changes

sections/api/primary/styled-component.js
```diff
@@ -17,30 +17,6 @@ const StyledComponent = () => md`
   next to the styles defined by the styled call.
   (e.g. \`<MyStyledComp className="bootstrap__btn" />\`)
 
-  #### .extend (_deprecated_)
```
kitten: I see this is removed now, but we probably want to keep it around, right? Otherwise our docs aren't valid for old versions and it might become confusing to some :)

probablyup:
Considering we’ve been warning people away from this API for a long time, I
thought it’d be better to just get rid of it yeah?

PR #290, opened 2018-08-22
Upgrade Next.js

timneutkens:
Upgrades Next.js to the latest version.

I've tried to do the most minimal upgrade, only changes are next.config, .babelrc and package.json.

The snapshots are failing, but I guess that's because of styled-components being upgraded.

If someone else could have a look at the tests that'd be great, as I haven't touched the website before


kitten:
Nice! I had a separate WIP PR for this, but it wasn't done properly, so let's start with yours and I can add on more upgrades later on.

The server-side rendering seems to be broken right now btw. Might also explain the failing image snapshots

timneutkens:
There's a super weird issue where it renders `[object Object]` 🤔

probablyup:
I think we might just not be babel 7 compatible https://github.com/styled-components/babel-plugin-styled-components/pull/134

timneutkens:
@probablyup I'm pretty sure styled-components works fine for a *lot* of Next.js users on Next.js 6+ though 🤔

probablyup:
Rolled into #293

timneutkens:
💯 Thanks @probablyup 🙏

PR #293, opened 2018-09-02
v4 prep & docs cleanup

### LINKED ISSUE #287: Update babel-plugin docs to hint at @babel/preset-typescript
Instead of setting up a TypeScript + Babel transpilation pipeline it's also possible to set up a pure Babel pipeline, as for other compile-to-JS languages now.

This has the added benefit of working the same as Flow or "vanilla" ESnext transpilation and is much easier to get right.

https://babeljs.io/docs/en/next/babel-preset-typescript.html

Related: https://github.com/styled-components/babel-plugin-styled-components/issues/41#issuecomment-410747606

probablyup:
Let's not merge until the first beta is cut

Closes #287


probablyup:
last thing to do here is update to the v4 beta when it gets cut and update usage of any of the removed syntaxes

mxstbr:
Cleaned up a little more but overall this is looking solid :+1:

mxstbr:
Something here broke the site, 500 error across the board

PR #300, opened 2018-09-06
Use mdx

mxstbr:
**TODO**

- [x] Setup next-mdx
- [x] Make it render our existing components
- [x] fix inline code
- [x] Move over all usage of the `md` component:
  - [x] sections
    - [x] advanced 
    - [x] api
    - [x] basics
    - [x] faqs
    - [x] tooling
    - [x] homepgae getting started
  - [x] pages (these two will be really hard/impossible since we do runtime md => react here)
    - [x] ecosystem.js
    - [x] releases.js

PR #301, opened 2018-09-06
Update and run prettier in precommit hook

mxstbr:
Just like we do for the main repo, not sure why we're not doing this here but we should!

Ref #300, this will also set us up to use prettier on the mdx files in the future.


components/Nav/Search.js
```diff
@@ -13,9 +13,7 @@ const Wrapper = styled.form`
 `
 
 const Input = styled.input`
-  ${resetInput}
-
-  flex: 0 0 auto;
+  ${resetInput} flex: 0 0 auto;
```
kachkaev: missing semicolon caused rows to merge (not just here)

mxstbr:
Yeah you're right @kachkaev we should base this off of #300. Closing in favor of a follow-up to #300.

mxstbr:
#305

PR #305, opened 2018-09-06
Implement prettier

mxstbr:
Replacement for #301 that's based on #300 to avoid tons of conflicts. (h/t @kachkaev) Until #300 is merged you'll also see those changes in this PR, which might make it little unwieldy.

Updates prettier and makes it run in the pre-commit hook to make sure all our code is consistently formatted!


kachkaev:
I can look at this once #300 is merged. Bullet point lists are now marked with `-` instead of `*`, which is cool. The only minor issue I see so far is to do with turning 

```
  ${resetInput}

  flex: 0 0 auto;
```

into 

```
 ${resetInput} flex: 0 0 auto;
```

in a few places inside styled-components css.

This happens because of no `;` after `${resetInput}` – should be easy to fix.

components/Layout.js
```diff
@@ -23,13 +23,13 @@ export const Content = styled.div`
...
     padding: ${rem(70)} ${rem(20)} ${rem(30)} ${rem(20)};
-    transform: translateX(${p => p.moveRight ? rem(sidebarWidth) : 0});
-  `)}
-
-  ${p => p.hero && css`
-    font-family: ${headerFont};
-    width: 75rem;
-  `}
+    transform: translateX(${p => (p.moveRight ? rem(sidebarWidth) : 0)});
+  `)} ${p =>
```
kachkaev: one more missing `;` here 🙂

components/Nav/NavButton.js
```diff
@@ -5,20 +5,19 @@ import { navbarHeight } from '../../utils/sizes'
...
-  ${p => p.active && css`
-    background: rgba(0, 0, 0, 0.07);
-  `}
-
-  &:focus {
+  ${p =>
+    p.active &&
+    css`
+      background: rgba(0, 0, 0, 0.07);
+    `} &:focus {
```
kachkaev: one more missing `;` here 🙂

pages/ecosystem.js
```diff
@@ -15,17 +16,28 @@ const Ecosystem = ({ readme, sidebarPages }) => (
...
-      This is an incomplete list of awesome things built with styled-components. If you have something to share, please add it to the <Link href="https://github.com/styled-components/awesome-styled-components" inline>awesome-styled-components</Link> repo on GitHub and it will automatically show up here!
+      This is an incomplete list of awesome things built with styled-components.
+      If you have something to share, please add it to the{' '}
+      <Link
+        href="https://github.com/styled-components/awesome-styled-components"
+        inline
+      >
+        awesome-styled-components
+      </Link>{' '}
+      repo on GitHub and it will automatically show up here!
```
kachkaev: Prettier does a not very good job of formatting jsx in 1.14 – see https://github.com/prettier/prettier/issues/4941

Perhaps using prettier from master on github is a better idea as it should not create such a mess and it will also support mdx

pages/index.js
```diff
@@ -149,13 +149,14 @@ const EditorContainer = styled.div`
 `
 
 const Editor = styled(LiveEditor)`
-  ${editorMixin}
-  height: 24rem;
+  ${editorMixin} height: 24rem;
```
kachkaev: one more missing `;` here 🙂  and there are a few more later on.

It might be worth fixing all this in the source and _then_ run prettier in just one commit on top of all other changes.

mxstbr:
- Upgraded to prettier master branch
- Ran prettier on all mdx files

This should be good to go! :tada:

probablyup:
@mxstbr needs a manual merge after the sidebar fix

mxstbr:
Shipping!

PR #158, opened 2017-11-01
Add/alogliadocs

morajabi:
Fix #85 

I've set up the search in the navbar. We are using [Algolia docsearhc](https://community.algolia.com/docsearch/) and the crawler config is live here https://github.com/algolia/docsearch-configs/blob/master/configs/styled-components.json
We can make PRs there to alter it. Also, the styling is the default one and we can customize the look by following the DocSearch documentation:
 (https://community.algolia.com/docsearch/documentation/) Which I tried and couldn't change it the way to be better and thought it is unnecessary at this point to put more time into it, but we can do that at any point.

PR #310, opened 2018-09-07
Initial work to add algolia to app

imbhargav5:
I requested algolia docsearch for config to add it out app. If we have a design I will update the UI and snapshots to match the designs. We will track it all here.


mxstbr:
The design from #158 looked pretty good, let's go with that?

imbhargav5:
@mxstbr I see one image related to the design for the input. Do we have any images that I can look at for reference as to how the input will look like on different devices?

probablyup:
@imbhargav5 I'd do the dropdown style for tablet (768) & up and then do a takeover style (full height & width with a close X) for 767 down

imbhargav5:
Sounds good.

imbhargav5:
@probablyup @mxstbr I am working on the mobile nav part right now. Do you have a screenshot/design for how the search would like on smaller screens? That would really help.

mxstbr:
😍 

Probably just the same thing but full-screen on mobile?

probablyup:
@imbhargav5 I was thinking something simple like this with a close "X" in the top right corner https://cdn.dribbble.com/users/186850/screenshots/4127824/reading_-_search_results.png

imbhargav5:
@probablyup I see. This helps. 😀

probablyup:
Can’t use css and a media query for the mobile look?

imbhargav5:
I think it's doable only if I change the structure of the navbar and mobileNavbar elements. A bit more than just media queries I would say. Will update PR in today or tomorrow.

imbhargav5:
So, inherently, the way Docsearch appends search results right beside the input is causing some layout issues but I made an ugly, very much work in progress modal for mobile. Please take a look and see if there are some issues that we can fix before I get into styling. @probablyup

probablyup:
Some notes so far:

1. It'd be great if the hover coloring etc could be adjusted to better match our website color scheme

2. We should have the search always be visible and live outside that collapse area for mobile

imbhargav5:
Ok. So for 2 should I put a search icon beside the collapse button?

probablyup:
I would figure out how the collapse works and then just move that whole component out of the collapse zone so it goes to the left of the collapse chevron

imbhargav5:
[Image from Gyazo](https://gyazo.com/2734ad5191397c026b114f724e9e8f19)

imbhargav5:
Let's go!!

sag1v:
@imbhargav5 That looks great!

PR #321, opened 2018-09-17
switch innerRef example to simple syntax highlighting

### LINKED ISSUE #320: Example on usage of innerRef is broken
screen shot 2018-09-16 at 1 28 59 pm
This example should focus the input on mouse enter, but it doesn't and throws the following error instead:

```
Uncaught TypeError: Cannot read property 'focus' of undefined
    at onMouseEnter (eval at z (react-live.es.js:495), <anonymous>:20:132)
    at Object.<anonymous> (react-dom.production.min.js:15)
    at Object.invokeGuardedCallback (react-dom.production.min.js:16)
    at Object.invokeGuardedCallbackAndCatchFirstError (react-dom.production.min.js:16)
    at P (react-dom.production.min.js:20)
    at I (react-dom.production.min.js:22)
    at R (react-dom.production.min.js:22)
    at Array.forEach (<anonymous>)
    at j (react-dom.production.min.js:21)
    at F (react-dom.production.min.js:24)
```

Seems like the obtained ref is undefined, hence this way of using innerRef doesn't work. Maybe it's a styled-components bug itself.

probablyup:
since we're using v4 for the website, the code can't run properly

closes #320

### DIFF
--- a/sections/api/old/inner-ref.md
+++ b/sections/api/old/inner-ref.md
@@ -14,7 +14,7 @@ To get a ref to the actual, wrapped DOM node, pass the callback to the `innerRef
 This example uses `innerRef` to save a ref to the styled input and focuses it once the user
 hovers over it.
 
-```react
+```jsx
 const Input = styled.input`
   padding: 0.5em;
   margin: 0.5em;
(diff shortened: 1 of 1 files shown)


sag1v:
Nice, I think we should iterate over all breaking changes that came with v4 and do the same changes in the docs.
(Maybe even create different docs versions per major?)
Is there a list of breaking changes somewhere?

PR #322, opened 2018-09-17
add documentation on selector patterns

probablyup:


sag1v:
Hi @probablyup, shouldn't the `&&&` "hack" get documented on this section as well?
I think not a lot of people know about it. 
[example1](https://github.com/styled-components/styled-components/issues/382#issuecomment-310921410), [example2](https://github.com/styled-components/styled-components/issues/860#issuecomment-337874677)

probablyup:
@sag1v good idea, added

PR #326, opened 2018-09-21
Move to fontawesome

imbhargav5:
@probablyup Let me know if this is good enough for us to get started with fontawesome (as we discussed). The idea is to use fontawesome for the search in the algolia PR as well. So let's merge this first?


imbhargav5:
@mxstbr Ah. Actually I didn't even think that far. I just followed the fontawesome guide for React. You are right. I used named imports and the test cases passed so I assumed all was well with respect to bundlesize. Can you think of any changes that I can do? I will take a look at the guides again.

probablyup:
Is their React package set up to do tree shaking? Once we upgrade to Next
7, the web pack 4 tree shaking stuff should start working.


On September 21, 2018 at 9:30:40 AM, Bhargav Ponnapalli (
notifications@github.com) wrote:

@mxstbr <https://github.com/mxstbr> Ah. Actually I didn't even think that
far. I just followed the fontawesome guide for React. You are right. I used
named imports and the test cases passed so I assumed all was well with
respect to bundlesize. Can you think of any changes that I can do? I will
take a look at the guides again.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/styled-components/styled-components-website/pull/326#issuecomment-423551694>,
or mute the thread

imbhargav5:
@mxstbr I just checked the bundle contents. The only things that got added to the bundle were github and medium icons. The other icons weren't added. We are only using those two now. The plan is to use search icon in the other PR. But yes there will be a bit of increase in the bundle size stemming from the base fontawesome logic code.

probablyup:
Also we should probably get rid of those snapshot tests at some point and just use something like https://percy.io/

We can talk to them about getting an open source plan

imbhargav5:
Awesome should I go ahead and merge this?

mxstbr:
That's awesome!! Thanks @imbhargav5

PR #356, opened 2018-10-11
Add deprecated label for old apis for algolia to show them better

imbhargav5:


stevemao:
Looks like algolia does not point to the dynamic environments so we can't really test this. But the change LGTM

stevemao:
@mxstbr Anything blocks this?

PR #360, opened 2018-10-16
Add note about attaching classes like always to FAQ

mxstbr:


mxstbr:
The example doesn't work, one sec

mxstbr:
Fixed, shipping this since I just posted a link to it on reddit! 😅

PR #363, opened 2018-10-17
fix cosmiconfig link

stevemao:
oops...


stevemao:
I completely missed that too. Sorry!

PR #368, opened 2018-10-22
Update react-native.md

necto68:
SAME import for react-native, isn't it?


probablyup:
Depends on what bundler you're using. Metro no longer does the aliasing by default. There's a docs note on that.

PR #370, opened 2018-10-24
Update babel-plugin documentation

breadadams:
+ Include extra minification/transpiling info from the babel plugin repo

In relation with https://github.com/styled-components/babel-plugin-styled-components/issues/146

### DIFF
--- a/sections/tooling/babel-plugin.md
+++ b/sections/tooling/babel-plugin.md
@@ -90,11 +90,11 @@ One example you might want to do this, is testing components with enzyme. While 
 > This option is turned on by default. If you experience mangled CSS
 > results, turn it off and open an issue please.
 
-This plugin minifies your styles in the tagged template literals, giving you big bundle size savings.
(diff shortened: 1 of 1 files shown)


sections/tooling/babel-plugin.md
```diff
@@ -124,12 +124,50 @@ It utilizes a babel helper to tag each styled component and library helper with
...
+Take a note that it will keep other template literals not related to styled-components as is:
+
+```js
+// Following will be converted:
+styled.div``
+keyframe``
+css``
+
+// But this will not be converted:
+`some text`
```
breadadams: @probablyup I noticed you (or maybe IDE formatter/prettier) _recently_ changed some of this example block's backticks to single quotes.

They _need_ to be backticks as that's what the codeblock is talking about; the fact that template strings without a "styled-component factory tag" aren't affected by the transpilation. Just fyi 🙂

probablyup: Prettier, yeah.

probablyup:
Thanks!

Issue #371, opened 2018-10-25
Missing example for withComponent


xehpuk:
In the documentation of `withComponent`, it reads:
> You can see it in action in the [Extending Styles](https://www.styled-components.com/docs/basics#extending-styles) section.

But there's not a single mention of it. I know it's getting replaced in favor of `as`, but it should still be documented.

probablyup:
Send a PR

PR #372, opened 2018-10-25
Put .withComponent into extending-styles.md (#371)

xehpuk:
Restored example for `.withComponent` as it was and is still available.


xehpuk:
> TypeError: Link.extend is not a function

Oh well, not sure how to handle this.

probablyup:
I'd rather just link to the two "previous API" pages than put full examples in there TBH. It'd get confusing for people otherwise if they're starting out with v4.

mxstbr:
Good call I think!

Issue #367, opened 2018-10-22
Unable to use website due to CORS problem


gshilin:
It happens on homepage and on docs too:
![image](https://user-images.githubusercontent.com/354111/47278706-b1a50d80-d5d3-11e8-982f-e62830967a33.png)

manpenaloza:
Just wanted to report this issue as it still seems exist.

<img width="1439" alt="sc-analytics-fetch-issue" src="https://user-images.githubusercontent.com/4241744/49871629-0c5e2700-fe17-11e8-9136-6b1b397d5d52.png">

Maybe that's not even be a CORS issue? Tried some ways to ping this https://sc-micro-analytics.now.sh/ analytics microservice but haven't succeeded. 

So whatever this microservice tracks: should it still work and has it ever had proper CORS settings to work with the styled-components.com domain?

manpenaloza:
ok, I'll go for it.

mxstbr:
Perfect, thank you! :pray:

mxstbr:
Fixed by the amazing @manpenaloza in #401!

PR #401, opened 2018-12-12
remove error-causing & deprecated analytics fetch

manpenaloza:
This pr exists to solve #367.

So far code changes in this pr at least solve the problem of getting the console error caused by the failing fetch call to the deprecated analytics microservice.

But to finish this properly: is there any need for analytics/tracking at all? 
If not, from current perspective the whole [track.js](https://github.com/styled-components/styled-components-website/blob/master/utils/track.js) file can be removed. From what i see, it's only purpose was to implement the analytics microservice. + Why I haven't removed track.js from technical perspective: the file is referenced in [next.config.js](https://github.com/styled-components/styled-components-website/blob/master/next.config.js#L37). I don't know exactly what happens to the website in production when I remove the `array push` statement in next.config.js. Don't wanna kill styled-components.com with this pr :) > anyone knows better than me?


mxstbr:
Try removing it entirely, and see if the deploy of the pushed commit breaks! 👍 I think we should be good to remove entirely.

manpenaloza:
thx @mxstbr, did so and the new deploy seems to work.

PR #403, opened 2018-12-18
Add example of how hoisting can help in a monorepo

janhoogeveen:
After fighting with lerna for 2 days, I think I found a way to use styled-components in a library package and import that package in my main application without warnings about duplicate imports.

I've added 2 examples of how to hoist styled-components (automatically and manually) to the documentation.


stevemao:
Hi @janhoogeveen, I think lerna hoists shared packages automatically. Also what's your issue with each package having it's own styled-components?

PR #434, opened 2019-03-01
Update SSR example to prevent memory leaks

thewazir:
We recently solved a memory leak and the solution was to use `try...finally` and call `sheet.seal()` in the `finally` block. The example for Next.js does this as well and I thought it would be better to make this more obvious by having it in the official documentation.


sections/advanced/server-side-rendering.md
```diff
@@ -53,6 +60,8 @@ You need to take this into account when adding the CSS string to your HTML outpu
 Alternatively the `ServerStyleSheet` instance also has a `getStyleElement()` method
 that returns an array of React elements.
 
+If rendering fails for any reason it's a good idea to use `try...finaly` to ensure that the `sheet` object will always be available for garbage collection.
```
mxstbr:
```suggestion
If rendering fails for any reason it's a good idea to use `try...finally` to ensure that the `sheet` object will always be available for garbage collection.
```

sections/advanced/server-side-rendering.md
```diff
@@ -25,8 +25,12 @@ import { renderToString } from 'react-dom/server'
 import { ServerStyleSheet } from 'styled-components'
 
 const sheet = new ServerStyleSheet()
-const html = renderToString(sheet.collectStyles(<YourApp />))
-const styleTags = sheet.getStyleTags() // or sheet.getStyleElement();
+try {
+  const html = renderToString(sheet.collectStyles(<YourApp />))
+  const styleTags = sheet.getStyleTags() // or sheet.getStyleElement();
+} finally {
```
probablyup: Let's make sure we keep a `catch` statement in here, the optional catch binding feature hasn't proliferated yet AFAIK.

probablyup: It's also worth a comment that you must not call `.seal()` until AFTER trying to use `sheet.getStyleElement()` or it'll throw a different error.

thewazir: @probablyup I added `catch` statements to both examples and added an explanation that `.seal()` should be called after `sheet.getStyleElement()` or `sheet.getStyleTags()`

imbhargav5:
@mxstbr I think this PR is ready.

PR #435, opened 2019-03-02
Should use props.size

SarahEW1206:
You have margin and padding taking their values from props.margin and props.padding respectively, so those styles are not being applied. They should take props.size as their value. :)


probablyup:
This actually is working but something on the prod website seems to be broken, weirdly. If you run the site locally it looks fine 🤔

Issue #439, opened 2019-03-13
Sort logos on homepage by website popularity


mxstbr:
Rather than alphabetically, we should sort the company logos on the homepage by approximate website popularity according to www.similarweb.com (global rank).

manpenaloza:
makes sense!

note: to avoid manual searching/sorting, just saw similarweb.com offers an api to get the data we'd need https://www.similarweb.com/corp/developer/ for sorting properly. it also offers a free api but that one has limitatations.
- 5 Results Per Metric
- 1 Month of Mobile App Data
- 3 Months of Web Traffic Data 
haven't found any info on whether they provide some more extended free api for open source projects. we can find out.

mxstbr:
Sounds great if you want to dig into that! :+1:

manpenaloza:
ok, I've sent a request to similarweb. let's if that's an option to at least automate sorting.

PR #441, opened 2019-03-14
sorted logos using SimilarWeb global rank as sorting indicator

manpenaloza:
So far I've implemented basic functionality requested in #439 but without the api request logic i mentioned in #439 (so far no feedback from SimilarWeb on my request for an api key). So did this searching/typing thingy to implement and test the feature organization taking in consideration an eventually upcoming SimilarWeb api implementation. But, data is valid, the component works as it should as far as I see and thus could be used for deployment if you want.

Some notes on the implementation:
- considering handling the global rank data receipt via api, I had to replace the React Fragment implementation with javascript array. Reason: a Fragment doesn't list its children (which would be all the logos), but using the api I'd need the children (logos) in order do to a proper element cloning appending the `similarWebGlobalRank` property.
- tried to modify only what is needed to achieve this, so rendering logic and base idea of a huge component including all logo components kept the same
- properties `key` and `similarWebGlobalRank` were added separately to the company logo component. but those props are highly company related. so compared this implementation it probably would make sense to build up company related datasets/objects which can be used to render company related stuff.

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -286,7 +286,7 @@ const Logos = [
     <BlueTomatoLogo />
   </CompanyLogo>,
 
-  <CompanyLogo bottom="-0.2rem" key="http://casperjs.org/" similarWebGlobalRank={1287463}>
+  <CompanyLogo bottom="-0.2rem" key="http://casper.com/" similarWebGlobalRank={33785}>
     <CasperLogo />
   </CompanyLogo>,
 
--- a/pages/index.js
+++ b/pages/index.js
@@ -237,127 +237,149 @@ const FooterContent = styled(Content)`
   `)};
 `;
 
-const Logos = () => {
+const SortedLogos = ({ children }) => {
+  // sorting logic: the more popular a website, the higher it gets listed
+  const similarWebGlobalRankSortedLogos = React.Children.toArray(children).sort(
+    (a, b) => a.props.similarWebGlobalRank - b.props.similarWebGlobalRank
+  );
+
   return (
-    <>
-      <CompanyLogo bottom="0.3rem" height="2.5rem">
-        <AirBnBLogo />
-      </CompanyLogo>
-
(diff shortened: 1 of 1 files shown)


pages/index.js
```diff
@@ -237,128 +237,150 @@ const FooterContent = styled(Content)`
...
+
+  <CompanyLogo bottom="-0.2rem" height="1.75rem" key="https://www.bloomberg.com" similarWebGlobalRank={507}>
+    <BloombergLogo />
+  </CompanyLogo>,
+
+  <CompanyLogo bottom="0.75rem" height="3rem" key="https://www.blue-tomato.com" similarWebGlobalRank={21322}>
+    <BlueTomatoLogo />
+  </CompanyLogo>,
+
+  <CompanyLogo bottom="-0.2rem" key="http://casperjs.org/" similarWebGlobalRank={1287463}>
```
mxstbr: This is actually `casper.com`, which should have a much higher rank!

manpenaloza: thx, updated it.

manpenaloza:
updated the branch with the `Casper` fix, ready for another review.

pages/index.js
```diff
@@ -237,128 +237,150 @@ const FooterContent = styled(Content)`
...
+  );
 
-      <CompanyLogo bottom="0.7rem" height="2.5rem">
-        <XingLogo />
-      </CompanyLogo>
-    </>
+  return (
+    <UsersWrapper>
+      {similarWebGlobalRankSortedLogos}
+      {similarWebGlobalRankSortedLogos}
```
mxstbr: Is there a reason they are rendered twice?

kitten: That's for the animation but previously the logos were in a component that was rendered twice, so this is correct 👌 there might be a smarter way to make this work though

manpenaloza: jap, saw it was rendered twice before, so I just kept it for now.

PR #445, opened 2019-03-26
Fix complex typescript example

koenpunt:
The example was using invalid syntax, and I updated it to be valid. I assumed the original author did mean to make it like this.

### DIFF
--- a/sections/api/typescript.md
+++ b/sections/api/typescript.md
@@ -157,11 +157,12 @@ import styled from 'styled-components'
 import Header, { Props as HeaderProps } from './Header'
 
 const Title =
-  (styled < { isActive: boolean }) &
-  (HeaderProps >
-    (({ isActive, ...rest }) => <Header {...rest} />)`
-  color: ${props => (props.isActive ? props.theme.primaryColor : props.theme.secondaryColor)}
-`)
+  styled <{ isActive: boolean } & HeaderProps>(
+    ({ isActive, ...rest }) => <Header {...rest} />
+  )`
+    color: ${(props) =>
+      props.isActive ? props.theme.primaryColor : props.theme.secondaryColor};
+`;
 ```
 
 This is the most complex example where we have specific properties for the styling of the component and pass


dan-kez:
Quick follow up - this change did not work for me. I had to do the following instead though I'm not sure if it's desired.
```TSX
const Title =
  styled <React.FC<{ isActive: boolean } & HeaderProps>>(
    ({ isActive, ...rest }) => <Header {...rest} />
  )`
    color: ${(props) =>
      props.isActive ? props.theme.primaryColor : props.theme.secondaryColor};
`;
```

Note the use of the `React.FC<...>`

sections/api/typescript.md
```diff
@@ -157,11 +157,12 @@ import styled from 'styled-components'
...
-    (({ isActive, ...rest }) => <Header {...rest} />)`
-  color: ${props => (props.isActive ? props.theme.primaryColor : props.theme.secondaryColor)}
-`)
+  styled <{ isActive: boolean } & HeaderProps>(
+    ({ isActive, ...rest }) => <Header {...rest} />
+  )`
+    color: ${(props) =>
+      props.isActive ? props.theme.primaryColor : props.theme.secondaryColor};
+`;
 ```
```
Igorbek: this particular example should work with even simpler syntax:

```ts
const Title = styled(Header)<{ isActive: boolean }>`
  color: ${(props) =>
    props.isActive ? props.theme.primaryColor : props.theme.secondaryColor};
`
```

probablyup:
@koenpunt can you make the requested changes when you have a moment?

sections/api/typescript.md
GuruM: This isn't the same thing though -- the example is trying to extract `isActive` from the props that are consumed by `Header`. In your example, `isActive` will still be a prop on `Header`

GuruM:
fwiw the neither the original example nor the fix has worked for me. There's been an open issue in the TS definition since Aug 2019: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37972

GuruM:
Actually, @dan-kez's example did work. Hurrah! Thanks!

Issue #460, opened 2019-06-03
Website Example Code Broken


cbravo:
## Steps to reproduce
visit https://www.styled-components.com/docs/basics#attaching-additional-props
## Expected Behavior
the two inputs should be different sizes based on what is written there.
## Actual Behavior
the two inputs are the exact same size and changing values has no effect.

adzshaf:
Hi, can I work on this issue? @mxstbr

probablyup:
Go for it

cbravo:
There is something really wholesome about this whole thread 😄

kelsonpw:
This issue is fixed in my PR.  Just waiting to get it merged.

PR #462, opened 2019-06-07
Fixing example - attaching additional props using .attr

kelsonpw:
Fixes the example on (https://www.styled-components.com/docs/basics#attaching-additional-props) #460 

Before:
image

After:
image


mxstbr:
The `margin: ${props => props.margin};` and `padding: ${props => props.padding}` should also both be `props => props.size`, right?

Otherwise this LGTM!

kelsonpw:
@mxstbr Yep good call.  It was a little confusing previously.  Updating both to `props => props.size`.

Let me know if anything else needs changing, but I think its good to go :)

sections/basics/attaching-additional-props.md
```diff
@@ -7,27 +7,26 @@ This way you can for example attach static props to an element, or pass a third-
 Here we render an `Input` component and attach some dynamic and static attributes to it:
 
 ```react
-const Input = styled.input.attrs(({ size }) => ({
+const Input = styled.input.attrs({
```
probablyup: This object-form attrs is actually deprecated, it'd be good to update it to function version instead:

```js
const Input = styled.input.attrs(props => ({ })`...`
```

kelsonpw: @probablyup 
I originally tried to write it that way, but the current version doesn't call `.attrs` as a function, it appears.

This site is currently using `"styled-components": "^4.0.0-beta.11", and from xmy testing it isnt until 4.2.0 where the function form works properly, but that breaks the app (.attrs is used in many places.)

The fix would be to update to 4.3.1 and update any `.attrs` app wide to use the function form right?  I would be happy to do the work and submit a PR.

PR #466, opened 2019-06-14
v5 beta prep

probablyup:


package.json
```diff
@@ -45,10 +45,12 @@
     "prismjs": "^1.6.0",
     "react": "^16.5.2",
     "react-dom": "^16.5.2",
+    "react-is": "^16.8.6",
     "react-live": "^2.1.2",
-    "styled-components": "^4.0.0-beta.11",
```
mxstbr: Love how we only upgrade styled-components when testing betas 🤣

sections/api/primary/index.md
```diff
@@ -14,4 +15,6 @@ import CssProp from './css-prop.md'
 
 <ThemeProvider />
 
+<StyleSheetManager />
```
mxstbr: Let's move this further down below, it really shouldn't be used very often.

sections/faqs/migration-v5.md
```diff
@@ -0,0 +1,22 @@
+## What do I need to do to migrate to v5?
+
+Ready for this?
+
+```sh
+npm install styled-components@beta react@^16.8 react-dom@^16.8 react-is@^16.8
+```
+
+_That's it._ 💥
```
mxstbr: We should probably mention that upgrading React (if you're on a lower version) isn't quite that simple?

sections/faqs/migration-v5.md
```diff
@@ -0,0 +1,22 @@
...
+
+styled-components v5 does not introduce any breaking public API changes, and adds the following:
```
mxstbr: Let's link to https://medium.com/styled-components/389747abd987, which should work even after publishing.

package.json
probablyup: Lol right? 😆

sections/faqs/migration-v5.md
probablyup: I think that's outside our concern, if you're doing a major React upgrade that's its own can of worms

PR #482, opened 2019-08-28
 Autofocus search textbox in mobile

SreedevSB:
Currently the user has to click twice to focus the search textbox in mobile. 
Fixed this problem by adding auto focus to textbox.


probablyup:
Thanks!

PR #437, opened 2019-03-11
Update TypeScript documentation

Ky6uk:
Fixes #392

PR #483, opened 2019-09-04
Fix invalid Typescript example of custom prop typings

remolueoend:
Hey everyone, as discussed [here](https://github.com/styled-components/styled-components-website/pull/437#issuecomment-527429146), I found an invalid typescript example and try to fix it in this PR. The current example contains syntax errors/invalid typescript. 

I further updated the contribution section of the README by replacing `npm` commands with `yarn`. There is no lock file for `npm` (package-lock.json), thus using `npm` for installing dependencies will cause a lot of unwanted side effects. 

Thanks for looking into this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,12 +22,12 @@ git clone https://github.com/styled-components/styled-components-website
 # Enter the repo
 cd styled-components-website
 # Install the dependencies
-npm install
+yarn install
 # Start local development
-npm run dev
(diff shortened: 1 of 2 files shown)


remolueoend:
I set up a small example to validate the proposed syntax: https://codesandbox.io/embed/objective-shape-t0vwb

PR #501, opened 2019-09-25
Upgrade to Next.js 9

Timer:
This upgrades the Styled Components website to Next.js 9.

What do we get?

- **Automatic static optimization**: Pages without `getInitialProps` (like the docs) are automatically rendered to _static HTML_ files! This means they'll be served from the edge as a CDN HIT.
- **API routes**: The custom server was deleted and the `/proxy/[asset]` route was changed into a Next.js API route (`/api/proxy/[asset]`). There is a `now.json` route that rewrites the old `/proxy/*` URLs for backwards compatibility.
- **Compatibility with Now v2**: This can be deployed as a Now v2 project and the `now.json` has been updated accordingly. This means faster and more stable deployments! Additionally, we're now leveraging CDN edge-caching instead of an in-memory Express cache for the 2x SSR pages.


Timer:
Closing in favor of https://github.com/styled-components/styled-components-website/pull/502

PR #502, opened 2019-09-25
Upgrade to Next.js 9

Timer:
This upgrades the Styled Components website to Next.js 9.

What do we get?

- **Automatic static optimization**: Pages without `getInitialProps` (like the docs) are automatically rendered to _static HTML_ files! This means they'll be served from the edge as a CDN HIT.
- **API routes**: The custom server was deleted and the `/proxy/[asset]` route was changed into a Next.js API route (`/api/proxy/[asset]`). There is a `now.json` route that rewrites the old `/proxy/*` URLs for backwards compatibility.
- **Compatibility with Now v2**: This can be deployed as a Now v2 project and the `now.json` has been updated accordingly. This means faster and more stable deployments! Additionally, we're now leveraging CDN edge-caching instead of an in-memory Express cache for the 2x SSR pages.


probablyup:
Woah that is smoking fast 👏

kitten:
This is awesome! Didn’t really have time to look into this myself ♥️😍 let’s totally get this merged ASAP!

pages/releases.js
```diff
@@ -40,8 +40,14 @@ const Releases = ({ releases, sidebarPages }) => (
   </DocsLayout>
 );
 
-Releases.getInitialProps = async () => {
+Releases.getInitialProps = async ({ res }) => {
   const releases = await getReleases();
+
+  if (res) {
+    // Revalidate this data every 30 seconds
+    res.setHeader('cache-control', 's-maxage=30,stale-while-revalidate');
```
kitten: I think we could totally increase this to a higher value. We don’t really need this page to be _too up-to-date_ 😅

Timer: Since the stale-while-revalidate header is present you'll never receive a CDN MISS, even after 30 seconds.

mxstbr:
Shipping, thanks so much for tackling this Timer, this is **amazing**!!!

PR #515, opened 2019-10-01
Added Culture Trip logo

ChristianRoper:

PR #517, opened 2019-10-02
CultureTrip logo Added

lakshya1604:


lakshya1604:
As per someones comment in the cultural logo theculturetrip.com should have been the right url .

mxstbr:
Already done in #515, thank you for the PR though!

PR #518, opened 2019-10-02
culturetrip logo gas been added

anjali-001:


mxstbr:
Already done in #515, thank you for the PR though!

Issue #275, opened 2018-07-03
Add showcase page


overra:
I [hear](https://twitter.com/mxstbr/status/1014169285852323841) @mxstbr would like a showcase page.

I'd like to help out. Do you have any specifics in mind?

mxstbr:
Awesome @overra! I think this should be a grid of screenshots of websites with their URLs and/or names, what do you think? From [these lists of users I compiled](https://mobile.twitter.com/mxstbr/status/1014405072061652993) I think we should highlight:

- vimeo.com (not sure this is built with s-c entirely, might only be parts? we should figure that out to know what to screenshot)
- target.com 
- patreon.com
- shop.lego.com
- ticketmaster.co.uk
- atlaskit.atlassian.com
- typeform.com
- coinbase.com
- spectrum.chat 🙈 
- vogue.de
- web.autocad.com
- tsm.gg
- invisionapp.com

I'd love to get a first version up and then iterate on/perfect it as we go along. Do you want to tackle that initial version?

jmiazga:
Is this a duplicate of #16?

mxstbr:
Kinda, my thinking is that a showcase is a lot easier since we don't need to get permission for screenshots vs logos. On top of that it's also nicer since you can immediately visually see what's built with styled-components 😅

overra:
@mxstbr Sure! I'll work on it this weekend :)

mxstbr:
@overra any news?

overra:
@mxstbr yep, sorry about that!

el1f:
Any news on this @overra? I'd love to take over if you don't have enough time to complete it!

el1f:
Done! We can close it :)

mxstbr:
Thank you @overra and @el1f! :100:

overra:
😬 can't believe it's been over a year. Sorry for dropping this. Thanks for picking it up @el1f!

PR #528, opened 2019-10-10
Showcase

el1f:
### Made the showcase section responsive (#275)
Didn't find any particular issue with @overra's implementation. He left quite a nice job when it comes to the page itself. I pretty much just made the limited changes necessary to make it responsive. The most disruptive change I made was centering the slideshow based on the screenshot itself rather than the block composed by screenshot + image to improve the looks of the page on smaller screens.

| Before centering        | After centering           |
| ------------- |:-------------:|
|before_slide_centering | after_slide_centering|

Tested the current implementation on the following physical devices:

- MacBook Pro 15"
- Ultrawide display (2560x1080)
- Huawei P20

Tested the current implementation on the following chrome simulated devices:

- Laptop (1440x900)
- Laptop (1280x800)
- Pixel 2* (411x731)
- iPhone 5* (320x568)
- iPhone 6* (375x667)
- iPhone 6+* (414x736)
- iPhone X* (375x812)
- iPad (768x1024)

*tested in portrait and landscape - [screenshots](https://github.com/styled-components/styled-components-website/files/3714952/screenshots.zip)


el1f:
Currently, the Navbar shows the showcase as its second item. IMO this should not be the case. I'd keep the left side of the Navbar to only show details strictly linked to the technical side of the library and push the more "PR and showcase" kind of pages to the right side of the navbar. Still trying to figure out whether they feel better on the right or the left side of the Algolia search bar, any opinions?
Still, this seems like material for a future PR as I'd rather not water down the contents of this one.

| Left | Right |
|---|---|
|image | image|

mxstbr:
I think the showcase is important enough to warrant being on the left side of the navbar next to the "Documentation" item, let's just keep it there for now.

Other than that, this looks great, thank you so much for tackling this! :100: One small visual bug and would you mind resolving those conflicts? Thank you.

Screenshot 2019-10-11 at 10 48 40

I think once we ship this we want to add some more sites to that page, and we also want to link to it from high up on the homepage as it's important to show that styled-components allows users to build nice websites. Want to open those two as issues so we don't forget to tackle them? :pray:

el1f:
Everything should be fixed now. Also, I've created #530 and #531 to keep track of the next steps :)

Issue #542, opened 2019-10-15
Update shop.lego.com to lego.com in showcase


mxstbr:
The website built with s-c recently moved from `shop.lego.com` to just lego.com

PR #551, opened 2019-10-15
Change lego address in manifest and rename related static file

sbielenica:


el1f:
#542 already dealt with this! Thanks for the contribution nonetheless!

Issue #540, opened 2019-10-15
Add Volkswagen logo to homepage


mxstbr:
https://www.volkswagen.de/de.html is using s-c

CodeSarthak:
Please close this issue as it has been resolved by #553

PR #553, opened 2019-10-15
added the volkswagen logo and changed the % of translate3d to show  a…

DorstM:
#540 I also changed translate3d so it would show all the logos and not reset at the middle of the slide


el1f:
Same as in #554 :)

Issue #536, opened 2019-10-12
Companies slider stutters after a certain amount of time


el1f:
The companies slider stops scrolling at a certain point and just loops back without showing all the logos. 

Preview of the issue:
https://youtu.be/j05qh7r4PIo

Deepam101:
Should it be the infinite running type without any gaps?

mxstbr:
Yes @Deepam101

Deepam101:
Please review #563 .If found correct please merge.
:)

el1f:
Great Job @Deepam101, now we can get back to adding even more logos guilt-free :)

PR #554, opened 2019-10-15
corrected the issue in which the slider was not showing all the logos

DorstM:
#536 
fixed it

### DIFF
--- a/components/UsersLogos/index.js
+++ b/components/UsersLogos/index.js
@@ -7,7 +7,7 @@ const slideAnimation = keyframes`
   }
 
   to{
-    transform:translate3d(-50%,0,0);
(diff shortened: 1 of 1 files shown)


el1f:
This is a step in the right direction, congrats! Still, now it has a gap when there are no more logos left to display. In my head, it should be an infinitely scrolling carousel instead of just stopping out of nowhere. Think you can invest a little more time on it?

Great job regardless!

el1f:
Well, this is still a step forward compared to the current state of things. Going to merge in order to then allow for anyone to build upon your contribution and make the transition less abrupt.
This will also allow for easier testing of all the PRs adding more logos.

Thanks!

Issue #544, opened 2019-10-15
Add BlaBlaCar to homepage logos


mxstbr:
All their country subsites use s-c, for example: https://www.blablacar.de/

anishkargaonkar:
May i take this up?

sai-chai:
This issue should be closed unless there's an issue with the PR that was merged

PR #556, opened 2019-10-18
Added BlaBlaCar to homepage logos

AlMilne:
Related issue: #544

Added  BlaBlaCar logo to homepage


companies-manifest.js
```diff
@@ -81,6 +82,11 @@ const companies = [
     similarWebGlobalRank: 114,
     logo: BBCLogo,
   },
+  {
+    key: 'https://www.blablacar.de',
+    similarWebGlobalRank: 13077,
```
mxstbr: Since bla bla car has a bunch of local websites, let's use their most widely used one, which appears to be `.fr`! https://www.similarweb.com/fr/website/blablacar.fr

```suggestion
    similarWebGlobalRank: 3803,
```

companies-manifest.js
```diff
@@ -81,6 +82,11 @@ const companies = [
     similarWebGlobalRank: 114,
     logo: BBCLogo,
   },
+  {
+    key: 'https://www.blablacar.de',
```
mxstbr: Since bla bla car has a bunch of local websites, let's use their most widely used one, which appears to be .fr! https://www.similarweb.com/fr/website/blablacar.fr

```suggestion
    key: 'https://www.blablacar.fr',
```

companies-manifest.js
AlMilne: Sounds good, have updated to use .fr version instead of .de

PR #565, opened 2019-10-25
fix the `.attrs` code example in the v4 migration docs

ali:
Fixes a mistake in the v4 migration docs. The "Before" and "After" code examples for migrating to the new `.attrs` API show the same code. This PR fixes the "Before" example so that it shows the old way that worked in v3/v4.


ali:
@probablyup Would you mind reviewing this PR when you get a chance?

probablyup:
This change is incorrect, the docs example is right.

ali:
Currently [both examples](https://github.com/styled-components/styled-components-website/pull/565/files#diff-ca13d7cf013aaaa8d14eccf53b358fd4L175-L194) show `.attrs` receiving a function that returns an object, and both functions are equivalent, so I'm otherwise not sure why one is 🚫and the other is ✅. My assumption was that the 🚫example should show `.attrs` receive an object containing functions as values (which worked in V3 but was later removed)

probablyup:
Returning an object that has a function key that consumes props is what is
deprecated

ali:
> Returning an object that has a function key that consumes props is what is deprecated

@probablyup Shouldn't the 🚫example demonstrate that usage, then? When I try that code, I don't see the warning in v4 about the deprecated API. Here's a codesandbox that shows both usage: https://codesandbox.io/s/nostalgic-banzai-wibtb

Issue #503, opened 2019-09-26
Document useTheme hook (v5)


mxstbr:
Reference PR: https://github.com/styled-components/styled-components/pull/2765

Should go in the API section and maybe be used in the `theming` section!

apollonian:
I can help with this one, @mxstbr! ⚡️

O45:
https://github.com/styled-components/styled-components-website/pull/230#issuecomment-364654261

yakovleva-tanya:
Hi @mxstbr, I'm looking for a good first issue and would like to contribute to this one if it's still available!

PR #569, opened 2019-10-31
#503 added useTheme hook documentation

arthurfalcao:
Related issue: #503 

Added useTheme hook in api helpers section


sections/api/helpers/use-theme.md
```diff
@@ -0,0 +1,16 @@
+### `useTheme`
```
probablyup: Needs a `| v5` tag in the header and then looks good to go

vzaidman:
any help needed with this one?

arthurfalcao:
> any help needed with this one?

just needs to be reviewed

sections/advanced/theming.md
```diff
@@ -97,30 +97,29 @@ If you ever need to use the current theme outside styled components (e.g. inside
...
   render() {
-    console.log('Current theme: ', this.props.theme);
+    console.log('Current theme: ', this.props.theme)
     // ...
   }
 }
 
-export default withTheme(MyComponent);
+export default withTheme(MyComponent)
 ```
```
probablyup, line 111: I'm not sure why you removed these docs, they're valid. `useTheme` is a separate API.

arthurfalcao: It was a @el1f 's suggestion 

> This is off to a great start @arthurfalcao! Thanks!
> Can you please try your hand at adding a couple of words about the hook on the theming section too though? There is a paragraph about accessing the theme using the `useContext` hook to get the theme, I think it could be integrated or even replaced by a mention of `useTheme`.

el1f: I'm probably mistaken but I feel like in the presence of the `useTheme` hook there aren't many use cases where the user's want to access the theme via the more verbose `useContext` hook. What do you think about it @probablyup ?

el1f: On second though I realize that users still using v4 could probably need the docs to keep mentioning the `useContext` method. Didn't really think about that possibility before... My bad.

el1f:
The only nitpick that I have is that IMO `useTheme` should come before `useContext` in the docs but that's probably quite subjective. Do you have any further comments on this @probablyup ?

probablyup (approved):
Thank you!

Issue #531, opened 2019-10-11
Link to the showcase page from the homepage


el1f:
The showcase page should be advertised on the homepage inviting visitors to check out other beautiful websites built with s-c.
I really like the way that next.js uses to do this, by having a smaller static showcase with some handpicked website screenshots to give a taste of what to expect from the page.

hairarrow:
Hey @el1f, if you haven't started on this, I'd love to tackle this.

PR #570, opened 2019-11-01
➕ Showcase link from the homepage

el1f:
Added a small showcase in the homepage to invite the user to visit the larger one as suggested in #531


mxstbr:
@el1f do you have Twitter?

PR #583, opened 2019-11-21
Update react-native.md

ralph-dev:
We need to clarify that Styling Children from the Parent Component is an Anti Pattern in React Native Styling and that 
* > {Style}
& > {Style}
${ChildComponent} > {Style}
Does not work in React Native. Had to find out from an external source.

Not Supported -> EX: https://www.geeksforgeeks.org/how-to-select-all-child-elements-recursively-using-css/


sections/basics/react-native.md
```diff
@@ -51,6 +51,9 @@ Some of the differences to the web-version are, that you cannot use the
 keyframes or global styles. We will also warn you if you use media queries or
 nest your CSS.
 
+> Note that React Native does not support Styling Children from the Parent
```
probablyup: Good start, I'd word it more like this:

Another difference is React Native does not support cascading styles, so when you write CSS for a React Native component it essentially becomes an inline style. The implication here is typical CSS selectors don't do anything, e.g. `* {}`, `> Text`, etc. It's recommended to make use of `ThemeProvider` to pass styles down to overcome this limitation.

ralph-dev: That's well worded as long as the implication is also defined in the paragraph as beginner developers may not be aware of the difference.

I think a provider is an interesting approach. If we want to suggest that I would like to see an example as to ensure developers are creating re-usable styles instead of implementing it incorrectly.

ralph-dev:
Any updates on this PR?

PR #592, opened 2019-12-04
Added auth0, artsy, brave and blablacar to showcase

el1f:


mxstbr:
Also, the images don't load for me in the preview 🤔

el1f:
> Also, the images don't load for me in the preview 🤔

Though I amended the commit before pushing. My bad.

PR #599, opened 2020-01-10
fix legacy attrs interface migration example

maxmalov:


ali:
I suggested this same correction in #565. I like that your PR goes a step further than mine and avoids destructuring the function arguments in both code examples, this makes the usage clearer to understand (IMO).

maxmalov:
Hmmm, why did you close it? It is clearly a typo in the current documentation, the both examples are actually equivalent. Instead it should explicitly define what pattern is deprecated 🚫 and how to replace it ✅.

Issue #576, opened 2019-11-04
Link to request a website addition


el1f:
We should add a button in the showcase section to create an issue that requests the addition of a new company and project.

PR #604, opened 2020-01-30
New showcase

### LINKED ISSUE #571: Add header to showcase page similar to homepage
The showcase insert on the homepage has some really nice phrasing:

Screenshot 2019-11-04 at 11 51 38

We should add some similar text to the dedicated /showcase page that explains what it is, something like:

----

# Showcase

styled-components is used by teams all around the world to create beautiful websites like these:

el1f:
Requires #603 to avoid showing incomplete data.

Refreshes the style of the showcase page and closes #571  and #576 

TODO before merging
- [x] Get feedback on the aesthetics 
- [x] Merge #603
- [x] Discuss tags for filtering
- [x] Clean up the code and split the `showcase` file into smaller components.


mxstbr:
This looks so much nicer! What's the "All:" dropdown for?

el1f:
It's currently taking space that could house a "tags" filter that'd filter the showcase to only show items of a specific category.
Thinking about it since the last commit though, I realized it makes little sense with the current structure that we have in the showcase since your "viewport" is limited to 3 items at a time and filtering the actual contents of the carousel wouldn't change much at all. I'm actually starting to consider adding an alternative mode, something like a simple grid view, to allow users to check more of the items in the showcase without spending hours on the page as currently the sites that don't stand at the very beginning of the carousel have little to no way to actually be seen by the user. In this case the filters make way more sense.

Any thoughts on that?

mxstbr:
I would like a grid view actually where you can see many items at once, that's how https://nextjs.org/showcase is laid out too! Let's ship this as-is and then move to a grid in a separate PR?

el1f:
Maybe I'm burning out, maybe I'm just plain dumb. Regardless, can any @styled-components/collaborators help out a fella here? Why are the styled-icons imports failing the Travis check? It only seems to happen on the showcase page and I can't reproduce the compilation error on my local environment. It only occurs on Travis...

Help?

StevenLangbroek:
Hey! Thought we could just work around the issue by importing directly from the files they're defined in but that just caused a boatload more issues. As I mentioned in my (now closed) MR, there's an issue in `styled-icon`, more people have reported intermittent and hard to reproduce import issues. I'll try to contribute there first 🙇

jacobwgillespie:
Have we seen this issue with the `@styled-icons/pack`-style imports?  I'm starting to wonder if it's the sheer number of files in the master `styled-icons` packages that's causing issues on some systems.  Styled Icons v10 is going to drop the `styled-icons/PACK/ICON` individual imports in favor of `@styled-icons/PACK/ICON` as we have hit a file count limit when publishing to NPM (`styled-icons/PACK` will still exist in v10).

@el1f you might want to pull the latest changes from `master` here as that branch has switched from importing `styled-icons` directly to only including the individual packs - change was in #601.  After pulling, you will have to update two imports in this PR (in `components/Slider/Navigation.js` and `components/Slider/ShowcaseBody.js`) like so:

https://github.com/jacobwgillespie/styled-components-website/commit/cc051ea77d58a402d58c20373057c32a169c0048

Given that this is an intermittent failure and there's many thousand small files in `styled-components`, switching to the smaller packages _might_ just resolve this.

StevenLangbroek:
@jacobwgillespie @egdbear #614's build just went through 🎉 thanks for the assist Jacob!

jacobwgillespie:
Oh awesome! 🎉

el1f:
Closed due to #614 merge.

PR #621, opened 2020-03-03
More logos

el1f:
image

A lot of people are joining our homepage but there isn't much space to let their presence shine right now. Borrowing (again) from Next.js I propose this little change to the homepage. What do you think about it @styled-components/website?


el1f:
Any @styled-components/website have any idea of the reason why my Travis builds keep failing when I commit?

PR #627, opened 2020-03-29
Add array join method to string reversal

adebiyial:
The reverse method returns an array, but what we want is a string. We need to join those arrays after reversal.

### DIFF
--- a/sections/basics/extending-styles.md
+++ b/sections/basics/extending-styles.md
@@ -75,7 +75,7 @@ const Button = styled.button`
   display: block;
 `;
 
-const ReversedButton = props => <Button {...props} children={props.children.split('').reverse()} />
+const ReversedButton = props => <Button {...props} children={props.children.split('').reverse().join('')} />
 
 render(
   <div>


sections/basics/extending-styles.md
```diff
@@ -75,7 +75,7 @@ const Button = styled.button`
   display: block;
 `;
 
-const ReversedButton = props => <Button {...props} children={props.children.split('').reverse()} />
+const ReversedButton = props => <Button {...props} children={props.children.split('').reverse().join('')} />
```
sweetliquid, line 78: I think we can extract it into utils:

```
const reveseString = str => str.split('').reverse().join('')
```

or use a library?

adebiyial: image

That's right. Does this work?

sweetliquid: It turned out to be just a doc, so just pretend I haven't said. lol

sandipsaha597 (approved):

PR #638, opened 2020-04-24
Add theming example with object notation

lukewlms:
Couldn't find this documented anywhere - we use exclusively object notation in our React Native project.  

This is fairly intuitive but should be explicitly documented.  Thanks!

### DIFF
--- a/sections/advanced/theming.md
+++ b/sections/advanced/theming.md
@@ -20,6 +20,7 @@ const Button = styled.button`
   border: 2px solid ${props => props.theme.main};
 `;
 
+
 // We are passing a default theme for Buttons that arent wrapped in the ThemeProvider
 Button.defaultProps = {
   theme: {
@@ -41,6 +42,14 @@ render(
     </ThemeProvider>
   </div>
 );
+```
+
+```
+// Example with object notation:
+const HighlightedSpan = Styled.Text(props => ({
+    color: props.theme.highlightTextColor,
+}));
+
 ```
 
 ### Function themes


sections/advanced/theming.md
```diff
@@ -20,6 +20,7 @@ const Button = styled.button`
   border: 2px solid ${props => props.theme.main};
 `;
 
+
```
el1f, line 23: Can you remove this newline please?

sections/advanced/theming.md
```diff
@@ -43,6 +44,14 @@ render(
 );
 ```
 
+```
+// Example with object notation:
+const HighlightedSpan = Styled.Text(props => ({
```
el1f, line 49: You should use `styled` instead of `Styled` to keep consistency with the rest of the documentation.
Also, what is `Text` in the context of this example? Perhaps it would be more fit to use a standard `span`?

lukewlms:
Thanks for the comments.  I don't have time to follow up further - this was difficult information to find so wanted to see if I could help surface it better.  Sounds like someone who understands the doc nuances better would be better able to follow up.

el1f:
Apologies for the nitpicking. I definitely see the value in your PR and will try to make time to give it a spin myself to see whether I can lay down the base for someone else to improve upon it just as you just did! Thanks for taking the time to point out the absence of this info though!

el1f:
Ok, this is embarrassing but it seems like the docs already have a paragraph on this [here](https://styled-components.com/docs/advanced#style-objects). Seems like we're both guilty of not scrolling to the very end of the page 😂

PR #639, opened 2020-04-30
Add SwissDev Jobs to showcase

Vrq:


companies-manifest.js
```diff
@@ -853,6 +854,26 @@ const companies = [
...
+    similarWebGlobalRank: 583400,
+    name: 'SwissDev Jobs',
+    logo: SwissDevJobsLogo,
+    style: {
+      bottom: '0.3rem',
+      height: '2.5rem',
+    },
+    projects: {
+      swissdevjobs: {
+        title: 'SwissDev Jobs transparent tech job board',
```
el1f, line 868: I think it would make more sense to only have the name of the website as the title and the description as the `description` field.

el1f:
It also seems like the build is failing. I'm currently reviewing from mobile so I can't investigate to a large degree of detail. Please check it out otherwise I'll try to see what's going wrong as soon as possible

Vrq:
I fixed some errors but now I don't know what might be the issue...

companies-manifest.js
```diff
@@ -853,6 +854,27 @@ const companies = [
       },
     },
   },
+  {
+    key: 'https://swissdevjobs.ch',
+    similarWebGlobalRank: 583400,
+    name: 'SwissDev Jobs',
+    logo: SwissDevJobsLogo,
+    style: {
```
el1f, line 862: This is weird but can you try committing with an empty object in the `style` key?

Vrq: Done, but still seems to be failing.

PR #640, opened 2020-05-02
Vertically center "Discover more" text 

pomber:
Hi! I noticed the text in the "Discover more" button wasn't centered and I couldn't resist to center it.

Before:
image

After:
image


el1f:
I genuinely appreciate this sprinkle of OCD. Thanks :D

PR #657, opened 2020-06-04
El1f/grid showcase

el1f:
Adds a grid layout for the styled-components showcase page. This allows users to check out all of our catalog of websites built with SC, increasing the likelihood that they're going to be seen.

image


mxstbr:
This is so good @el1f!! :fire: A couple small things:

- I would remove the slideshow-view entirely and only have the grid view, it's so much better
- I would make the default grid show rows of 2 screenshots next to each other (instead of 3) so they're all the screenshots are a bit bigger
- I would add a persistent indicator of what the site is below the screenshot, rather than only showing that on hover. Maybe something like this:
  Screenshot 2020-06-04 at 21 42 27
  Depending on how big the screenshots are with only two next to each other, we might even be able to show the name + the URL down there? 🤔 

What do you think?

kitten:
I disagree on all points except for showing the name of the page. Often it’s not relevant given that most of these sites reflect that brand’s image and identity.

As long as there are appropriate alt-tags or the text is still screen-reader accessible, I’d skip it.

el1f:
Thanks for the proposals guys! Here's my latest iteration. I've converted the display mode switcher into a density picker to allow users on large monitors to pick between seeing 2 or 3 sites per row.

| Regular | Dense |
| --- | --- |
| image | image |

It now also supports different sorting logic. Currently, you can only pick between Alphabetical or Popularity ordering but I want to add timestamps to the addition of the websites to the Manifest in order to show them sorted by the newest addition to oldest.

**TODO before landing this:**
Still have to clear a lot of unused code left over by the slideshow view.

mxstbr:
> Often it’s not relevant given that most of these sites reflect that brand’s image and identity.

To me, it's not at all obvious at first glance which brands are featured for most screenshots (the occasional exception like Airbnb proving the rule):

Screenshot 2020-06-23 at 10 15 38

I can't even read most of the logos in the navbars because they're too small 😅 I would prefer if we included the names but don't feel that strongly about it (7/10 maybe)

PR #667, opened 2020-06-19
feat(showcase): add spotify

### LINKED ISSUE #661: Add spotify.com to the showcase
The main https://www.spotify.com landing page uses styled-components! :fire:

junagao:
resolves #661


junagao:
Hey @mxstbr,

I did run `yarn test -u`, all tests passed but no snapshots were updated. Could you provide more information on how to proceed in this case?

Thank you ☺️

kitten:
@junagao If I remember correctly there's a test that snapshots the `LiveEditor` which also (erroneously) snapshots the grammar for Prism, which it probably shouldn't. This contains some RegEx strings which are stringified differently depending on the Node version (again, not sure why 😓). So you can just reset `test/components/LiveEdit.spec.js`'s snapshot based on `master`. We really need to fix up this test 😅

junagao:
Hi @kitten,

Sorry, this is my first contribution here. I'm not sure I understand. How can I reset `test/components/LiveEdit.spec.js's snapshot based on master? Could you please provide any tips?

Thank you.

el1f:
@junago You should be able to quickly fix this by just copying the `test/components/LiveEdit.spec.js` from master onto your tests after running `yarn test -u `

junagao:
Hi @el1f , I did as you said but it's actually the exact same snapshot. There wasn't any change inside the `test/components/LiveEdit.spec.js` comparing between the master and the feat/add-spotify-to-showcase branch.

Maybe if I install another Node version it will give me the correct updated snapshot. I will try it.

junagao:
It worked with an older Node version 👍

el1f:
That's great! That's actually what I'm also doing right now. I'll also look into how to solve this issue with the tests but I'm not that good when it comes to fixing tests so I can't assure much :P

el1f:
Also, @junagao, I have just a tiny nitpick:
Can you set the title of the project to "Spotify Website" as just saying Spotify honestly makes me think about the Spotify desktop or web app.
It may be just me but it would be nice to avoid any possible confusion 😊

junagao:
@el1f sure, will do. :)

PR #668, opened 2020-06-26
Add imdb.com to showcase

### LINKED ISSUE #660: Add imdb.com to showcase
imdb.com is using styled-components! 😍

koalamango:
resolves #660


koalamango:
@mxstbr I've reverted LiveEdit.spec.js.snap as my changes shouldn't affect that snapshot.

For info, when I checkout master without any changes, if I run ```npm test``` or ```yarn test``` then the test fail. If I run ```yarn test -u``` to update snapshots then test pass locally but fails on Travis-CI.

Don't know where to start with that one!  🤪

PR #669, opened 2020-06-26
Fix round brackets on typescript example

mglavall:
The round brackets of the snippets were wrong.


mglavall:
What's the next step to be able to deploy this?

kokushkin (dismissed):
It might be not how it works, but definitely much better than was before.

PR #675, opened 2020-07-13
Update z-index and add cursor on small showcase hover (fix #674)

### LINKED ISSUE #674: Update z-index on small showcase site hover to prevent clipping.
Hey Styled Components Friends! 👋

I noticed that the "Small Showcase" on the home page has a couple of issues when hovered. 

clipped-showcase

1. The first is that when a site is hovered, the label is sometimes clipped by the site thumbnail resting in front of it.

2. There isn't a cursor/ pointer shown when a site is hovered. While the majority of users probably know they can click through, others may not and think that their only option to see or learn more is by clicking the discover more button below. 

That said, I've created a pull request that basically applies the opposite z-index and a cursor when hovered. This way the hovered site will be move to the top and will be readable, etc. I'll reference in the comments below. 

Thanks in advance for your time and support.

justinjunodev:
My proposal for fixing the issues described and outlined in: [Issue #674](https://github.com/styled-components/styled-components-website/issues/674).


el1f:
Love It from my phone! Will test later on my laptop and let you know about any other nitpicks :P
Thanks for your help for now man! :D

justinjunodev:
No worries, @el1f! Excited to see this live. Appreciate the feedback and collaboration. 👍

PR #681, opened 2020-07-20
Resized homely showcase thumbnail

Morganjackson:


probablyup:
@Morganjackson this is 150kb larger than the original, can you run it through a lossless shrinker?

el1f:
Hey @Morganjackson, can you please update your PR with this source? It's about 51% lighter and it should allow us to close this PR right after that :)

homely com au

Morganjackson:
@el1f sorry for the delay should be good to go now!

el1f (approved):
While it's still slightly bigger than the original it is by a very small margin so I'll go ahead and approve it. Thanks for your help @Morganjackson!

PR #682, opened 2020-07-20
Update showcase.js to use headerFont

Morganjackson:
Without this, the fonts look broken unless the user has **Avenir Next** installed.

**Before:**
before

**After:**

after


probablyup (dismissed):
This code needs a general refactor but in the meantime happy to take the patch, thank you

probablyup (approved):

PR #683, opened 2020-07-20
Updated outbound links to use standard anchor tags.

Morganjackson:
See https://github.com/vercel/next.js/pull/8231/files/97f86eec4c400af05114d9632bfdfce996f1e955 for details.


probablyup:
@Morganjackson perhaps we can take this opportunity to make a smarter link component instead? Basically inspect the `href` and if it contains a full path instead of a relative one, don't wrap with next/link

Issue #701, opened 2020-07-29
Add zillow.com to showcase


mxstbr:
Apparently https://www.zillow.com/ is built with styled-components!! :scream:

dkbearsong:
I have submitted a pull request that covers this issue.

PR #710, opened 2020-08-14
Gizmodo lifehacker redbull

brnt-toast:
this branch is built off the #701 branch(add-zillow)...
I'm not really confident if this was the right approach and that i should have only sent one PR. I am still a bit green, and have my sea legs. If someone could critique that would be great and i would appreciate the help


probablyup:
Covered by recently merged PRs, thanks!

Issue #699, opened 2020-07-29
Add Gizmodo to the showcase


mxstbr:
Turns out https://gizmodo.com/ is (at least the homepage) built entirely with styled-components! :fire:

bovas85:
working on this 🔥

bovas85:
PR https://github.com/styled-components/styled-components-website/pull/726

Issue #698, opened 2020-07-29
Add Lifehacker to the showcase


mxstbr:
Turns out https://lifehacker.com/ partially uses styled-components for their website! :scream: We should add them to the showcase and companies list.

PR #713, opened 2020-08-17
added zillow, gizmodo, giphy, and lifehacker to the showcases

dkbearsong:
updated comapnies-manifest.js to include entries for the four mentioned websites, added screenshots for each website, and added companlogos.js to also include their logos.

This addresses issues #698 #699 #700 and #701


probablyup:
@dkbearsong gizmodo already got added so I'm going to close this one, don't need links for every property. Thanks!

probablyup:
@dkbearsong if you want to reduce this down to just zillow though we'll take that

PR #720, opened 2020-08-31
clarify how multiple layers of .attrs work

vzaidman:
clarify how multiple layers of .attrs work to prevent confusions like in this issue: https://github.com/styled-components/styled-components/issues/3178

### DIFF
--- a/sections/basics/attaching-additional-props.md
+++ b/sections/basics/attaching-additional-props.md
@@ -9,7 +9,7 @@ Here we render an `Input` component and attach some dynamic and static attribute
 ```react
 const Input = styled.input.attrs(props => ({
   // we can define static props
-  type: "password",
+  type: "text",
 
   // or we can define dynamic ones
   size: props.size || "1em",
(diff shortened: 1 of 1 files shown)


sections/basics/attaching-additional-props.md
```diff
@@ -34,3 +34,38 @@ render(
 ```
 
 As you can see, we get access to our newly created props in the interpolations, and the `type` attribute is passed down to the element.
+
+### Overriding .attrs
+Notice that when wrapping styled components, `.attrs` are applied from the innermost styled component to the uppermost styled component.
```
probablyup, line 39:
```suggestion
Notice that when wrapping styled components, `.attrs` are applied from the innermost styled component to the outermost styled component.
```

sections/basics/attaching-additional-props.md
```diff
@@ -34,3 +34,38 @@ render(
...
+
+This allows each level to **override** lower levels of `.attrs`, similarly to how css properties override css properties of inner components.
```
probablyup, line 41:
```suggestion
This allows each wrapper to **override** nested uses of `.attrs`, similarly to how css properties defined later in a stylesheet override previous declarations.
```

sections/basics/attaching-additional-props.md
```diff
@@ -34,3 +34,38 @@ render(
...
+
+Here `Input`'s `.attrs` are applied first, and then `PasswordInput`'s `.attrs`:
```
probablyup, line 43:
```suggestion
`Input`'s `.attrs` are applied first, and then `PasswordInput`'s `.attrs`:
```

probablyup (changes requested):

vzaidman:
took all suggestions. thanks!

probablyup (approved):
Thank you for the improvement!

Issue #716, opened 2020-08-24
Typo: "Overidden" instead of overriden


sk22:
A button in https://github.com/styled-components/styled-components-website/blob/master/sections/advanced/theming.md says "Overidden" instead of "Overridden."

dhanush1909:
Hi, can I work on this?

PR #724, opened 2020-09-15
Fix typo in theming.md file

joe-orav:
#716

Changed "Overidden" to "Overridden" in the theming.md file

### DIFF
--- a/sections/advanced/theming.md
+++ b/sections/advanced/theming.md
@@ -97,16 +97,16 @@ If you ever need to use the current theme outside styled components (e.g. inside
 the `withTheme` higher order component.
 
 ```jsx
-import { withTheme } from 'styled-components';
+import { withTheme } from 'styled-components'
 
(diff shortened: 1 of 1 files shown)


sk22 (changes requested):
unintended code style change next to typo fix :/

joe-orav:
> unintended code style change next to typo fix :/

Is this regarding the semi-colons? If so, the package.json file specifies a "--no-semi" option for prettier when commiting changes in .md files. I'm not able to override that unless that needs to be removed?

sk22:
> > unintended code style change next to typo fix :/
> 
> Is this regarding the semi-colons? If so, the package.json file specifies a "--no-semi" option for prettier when commiting changes in .md files. I'm not able to override that unless that needs to be removed?

Oh huh, well then I guess it should be fine then

sk22 (approved):

probablyup:
Fixed in #727

PR #736, opened 2020-10-12
🐛 Image component main tag creates multiple mains per page

### LINKED ISSUE #715: Semantic errors in the Showcase body
There are 3 main tags on the page.  They can be found in the figure element, which is nested below the nav tag in the showcase body div. Semantically there can only be one main tag. It makes sense for the for the screenshot(16:9) in the showcase body which is mostly correct in hierarchy as well. The thumbnails, however, not so much.

el1f:
closes #715 
will be additionally resolved by #657


probablyup (approved):

probablyup (approved):
Thank you for catching this 👍🏼

PR #741, opened 2020-10-28
Fix mobile navbar layout

sweetliquid:
# Before

image

# After

image


sweetliquid:
Sory, I forgot autosquash.

PR #743, opened 2020-12-03
Improve the docs for how to use ampersands

probablyup:
Documentation improvements to solve https://github.com/styled-components/styled-components/issues/3334


sections/basics/coming-from-css.md
```diff
@@ -103,46 +103,123 @@ and comparisons with alternatives, in [Styled Components: To Use or Not to Use?]
...
 
-The ampersand (`&`) can be used to refer back to the main component. Here are some more examples of its potential usage:
+Through this preprocessing, styled-components supports some advanced selector patterns:
 
-```react
-const Thing = styled.div.attrs((/* props */) => ({ tabIndex: 0 }))`
-  color: blue;
+* `&` a single ampersand refers to **all instances** of the component; it is used for applying broad overrides:
+  
+  ```react
```
kitten, line 110: This indentation may render a little oddly; so I'll double check that. Apart from this, I'm wondering whether this is a little long. It may work to just describe the operators individually then list them out in a larger code block 🤔 

Also I'm getting a bit lost here but component-wide interpolations are called out, but I'm not sure whether they're there explicitly, e.g. `${Input}`?

PR #747, opened 2020-12-18
Doc updates for upcoming DT changes

orta:
Re the upcoming changes in  https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49914

I'll probably merge ^ today, so ideally this could go through pretty soon

PR #749, opened 2020-12-18
Use yarn instead of npm in installation instructions

tywmick:
So no one accidentally creates a `package-lock.json` (which would conflict with `yarn.lock`).


Gianluska (approved):
All good yet.

tywmick:
Oh, looks like this was fixed in #747. 👍🏼

PR #781, opened 2021-04-22
fixed run-on sentences in docs

KennethKinLum:


sections/basics/coming-from-css.md
```diff
@@ -86,7 +86,7 @@ Instead of:
 
 ```jsx
 const Wrapper = ({ message }) => {
-  // WARNING: THIS IS VERY VERY BAD AND SLOW, DO NOT DO THIS!!!
+  // WARNING: THIS IS VERY VERY BAD AND SLOW. DO NOT DO THIS!!!
```
davilima6, line 89:
```suggestion
  // WARNING: THIS IS VERY VERY BAD AND SLOW. DO NOT DO THIS!
```

sections/basics/coming-from-css.md
```diff
@@ -68,7 +68,7 @@ Note that we added a "Styled" prefix to `StyledCounter` so that the React compon
 
 ### Define Styled Components outside of the render method
 
-It is important to define your styled components outside of the render method, otherwise it will be recreated on every single render pass. Defining a styled component within the render method will thwart caching and drastically slow down rendering speed, and should be avoided.
+It is important to define your styled components outside of the render method; otherwise it will be recreated on every single render pass. Defining a styled component within the render method will thwart caching and drastically slow down rendering speed, and should be avoided.
```
kitten, line 71: It's a conjunctive, so we can treat it as an entirely separate sentence and follow it up with a comma

```suggestion
It is important to define your styled components outside of the render method. Otherwise, it will be recreated on every single render pass. Defining a styled component within the render method will thwart caching and drastically slow down rendering speed, and should be avoided.
```

PR #783, opened 2021-05-02
Add SwissDevJobs to showcase

Vrq:


Vrq:
I cannot access the build logs, what was the issue?

PR #789, opened 2021-06-16
Modified time animation logos

horacioss:
Hi everyone. The speed of the logo animation movement causes vertigo and headache. Since the sections are moving in opposite directions, it appears that they will translate to faster.

https://user-images.githubusercontent.com/26256117/122799190-6ca76c80-d28f-11eb-9334-e783d95622a1.mp4


probablyup (approved):
Looks great, thank you!

PR #792, opened 2021-07-09
Fix snapshot example

JuhQ:
The snapshot example contained color: green; even though the example code contains color: red;


probablyup (approved):
Thank you :)

PR #793, opened 2021-07-13
Updating theme.md to include useTheme hook

AleksandrukTad:
Not sure in which  version was useTheme added my guts tell me that it was v4 but if I am wrong please correct me.


probablyup (approved):
Thank you!

probablyup:
@AleksandrukTad can you add this to the general API documentation as well?

probablyup:
I missed that there was an earlier PR that includes my requested change, so we're going to merge that one. Thanks for contributing! 569

PR #795, opened 2021-07-21
Fix syntax and formatting in styled-theming.md

akd-io:
Fixed a couple of syntax errors in font-family strings, as well as a bit of formatting regarding self-closing JSX elements.


probablyup (approved):
Thank you!

PR #847, opened 2022-03-01
replace react-live with react-runner

nihgwu:
better performance and SSR support


probablyup (approved):
Looks to work well and confirmed the bundle savings. Thanks for taking this on!

nihgwu:
@probablyup I'm just about to update the package, will create another PR

PR #849, opened 2022-03-06
Change 'code-splitted' to 'code-split'

arye-eidelman:
Context: "Keyframes are lazily injected when they're used, which is how they can be code-splitted, so you have to use the `css` helper for shared style fragments"


probablyup (approved):
Makes sense, thanks!

PR #855, opened 2022-03-25
update react-runner to v1.0

nihgwu:
There is no change since `rc.2`, but I changed the example in home page, so we can only edit the style, previously if we remove some code, the render part will be shown, which I think is not intended, as I can see if we don't edit the code, the render part is not editable, so I changed to simply hide it from editing, correct me if I'm wrong on that


nihgwu:
followup of https://github.com/styled-components/styled-components-website/pull/847

PR #863, opened 2022-05-27
docs: Add new react native declarations file

danieltvaz:
I've got stuck trying to type my theme props passed by ThemeProvider, then I found myself i've extending DefaultTheme from styled-components module, not for styled-components/native one. I think this little change can speed up other developers to setup the provider.


probablyup (approved):
Yes definitely thank you

PR #864, opened 2022-05-31
fix: remove typings from production dependencies

rgomesrn:
typings should not be included in production dependencies


probablyup (approved):
Thanks :)

PR #887, opened 2023-01-03
Add target _blank on external links for improve user experience

iamdevmarcos:
Added target _blank on external links for improve user experience and accesibility.
This way the user will not leave the website after clicking on a social media link, thus improving their experience.

### DIFF
--- a/components/Link.js
+++ b/components/Link.js
@@ -33,7 +33,7 @@ export const InlineLink = styled.a.attrs((/* props */) => ({
   }
 `;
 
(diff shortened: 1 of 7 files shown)


components/Link.js
```diff
@@ -46,9 +46,14 @@ const Link = ({ ['aria-label']: ariaLabel, children, className, inline, unstyled
     dataAttrs = { 'data-white': white };
   }
 
+  let target;
+  if (isExternal) {
```
probablyup, line 50: What if instead of an explicit `isExternal` flag we just looked at `href` and determined if it's a relative path or not?

PR #905, opened 2023-04-09
changed installation instructions

ranidu-harshana:


probablyup:
The longform syntax is intentional because it’s more obvious, but it does seem like the save switch can be dropped

ranidu-harshana:
> The longform syntax is intentional because it’s more obvious, but it does seem like the save switch can be dropped

ok. may I do it or close it

probablyup:
@ranidu-harshana if you could just put `i` back to `install` then that would be good thanks

ranidu-harshana:
@probablyup Done

probablyup (approved):
Thanks!

PR #906, opened 2023-04-09
clarify typescript documentation examples for react-native

Illu:
Remove some confusion around the code snippets for setting up the theme interface with React-Native:

- Add missing import for the declaration merging example
- ~~Replace the `ThemeInterface` import with the `DefaultTheme` that was augmented in the previous step. The `ThemeInterface` and the `theme.ts` file from which it is imported are never mentioned before this step.~~
- Remove unnecessary code snippet

### DIFF
--- a/sections/api/typescript.md
+++ b/sections/api/typescript.md
@@ -46,6 +46,8 @@ declare module 'styled-components' {
 React-Native:
 
 ```ts
+import 'styled-components/native'
+
 declare module 'styled-components/native' {
   export interface DefaultTheme {
     borderRadius: string;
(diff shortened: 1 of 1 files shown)


sections/api/typescript.md
```diff
@@ -85,14 +87,13 @@ React-Native:
...
-
-import ThemeInterface from "./theme";
+import DefaultTheme from "styled-components/native";
 
 const {
   default: styled,
   css,
   ThemeProvider
-} = styledComponents as styledComponents.ReactNativeThemedStyledComponentsModule<ThemeInterface>;
+} = styledComponents as styledComponents.ReactNativeThemedStyledComponentsModule<DefaultTheme>;
```
probablyup, line 96: I'm not clear on why we need this code...

the `*.d.ts` file augmenting `styled-components/native` should be enough

Illu: Tested it and this whole code snippet seems indeed useless, the declaration file is all we need.
I removed it from the page altogether.

probablyup (approved):
Thank you!

Issue #904, opened 2023-04-06
Increase Accessibility


brnt-toast:
1. add names to buttons
2. form elements do not have labels
3. links do not have names

Mish2j:
Hi there, I would like to work on this.

Knikkey:
If this issue is still open, I'd like to contribute.

Cyanhead:
Hi👋🏽! Is this still open? I'd like to contribute.

kwconrad:
Has this issue been closed? If not, I'd love to help out.

tharikashree:
Would you assign this to me?

JorgeOteiza:
Hello! Is this issue still open?

rephlax:
Hello, would I be able to help with this?

PR #910, opened 2023-04-25
Increase Accessibility #904

Mish2j:
what needs to be done:
1. add names to buttons
2. form elements do not have labels
3. links do not have names

#1 and #3 have been completed. For #2 there is no elegant way to add labels on form elements, since the form element (textarea) is a child element of the "LiveEditor" component imported from "react-live-runner".

### DIFF
--- a/components/Nav/Navbar.js
+++ b/components/Nav/Navbar.js
@@ -173,7 +173,7 @@ class Navbar extends PureComponent {
       <Wrapper $transparent={transparent}>
         <NormalNavbar>
           <StartWrapper>
-            <LogoLink ariaLabel="Home Page">
+            <LogoLink ariaLabel="Styled Components Home Page">
               <Logo />
             </LogoLink>
             <NavLinks />
--- a/components/Nav/Navbar.js
+++ b/components/Nav/Navbar.js
@@ -136,7 +136,7 @@ class ModalContainer extends PureComponent {
     return (
       <>
         <AlgoliaModalHeader $isOpen={this.props.isOpen}>
-          <button onClick={this.onCloseButtonClick}>
+          <button onClick={this.onCloseButtonClick} aria-label="Close Search Bar">
             <StyledModalCloseIcon />
           </button>
         </AlgoliaModalHeader>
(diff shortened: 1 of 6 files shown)


components/Nav/Navbar.js
```diff
@@ -150,9 +150,10 @@ class ModalContainer extends PureComponent {
   }
 }
 
-const LogoLink = styled(Link).attrs((/* props */) => ({
+const LogoLink = styled(Link).attrs(({ ariaLabel }) => ({
```
probablyup, line 153: I think you can just pass `aria-label` directly on `LogoLink` instead of using the `attrs` indirection:

```tsx
<LogoLink aria-label="Styled Components Home Page">
```

components/Nav/Social.js
```diff
@@ -11,8 +11,9 @@ const Wrapper = styled.nav`
   flex: 1 1 auto;
 `;
 
-const SocialLink = styled(Link).attrs((/* props */) => ({
+const SocialLink = styled(Link).attrs(props => ({
   unstyled: true,
+  'aria-label': props.ariaLabel,
```
probablyup, line 16: Same here

probablyup:
This is looking pretty good to me, the only comment I have is we might want to just use `title` instead of `aria-label` so you get the nice hovertext as well for non-screen readers

Mish2j:
> This is looking pretty good to me, the only comment I have is we might want to just use `title` instead of `aria-label` so you get the nice hovertext as well for non-screen readers

Hi @probablyup, thanks for the feedback. I'll work on this in the next few days... 
Regarding the `title` attribute, I think it would be better to have both `title` and `aria-label` since `title` is not always reliable as stated on [W3.org](https://www.w3.org/WAI/tutorials/forms/labels/#using-the-title-attribute)

> The title attribute can also be used to identify form controls. This approach is generally less reliable and not recommended because some screen readers and assistive technologies do not interpret the title attribute as a replacement for the label element, possibly because the title attribute is often used to provide non-essential information. The information of the title attribute is shown to visual users as a tool tip when hovering over the form field with the mouse. 

Let me know what you think...

PR #929, opened 2023-07-03
Fix attrs examples

benbryant0:
I noticed a few minor things with the `.attrs` examples in the docs.

First, fixed a tiny syntax error in the example [here](https://styled-components.com/docs/basics#overriding-.attrs) that caused an error to show up in the page:
image

Next, the generic type parameters seem to be in the wrong position, as the examples all give me type errors when I copy them (SC 6.0.2 & TS 5.1.6):
image
I could easily be wrong here as I was actually looking at the docs for examples of correct generics with `.attrs` 😅, but moving the param seems to resolve the error and the props still get typed properly when using the attr-wrapped component.

### DIFF
--- a/sections/api/primary/styled-component.mdx
+++ b/sections/api/primary/styled-component.mdx
@@ -47,10 +47,10 @@ Returns another `StyledComponent`.
 ```react
 // import styled from 'styled-components'
 
-const Input = styled.input.attrs(props => ({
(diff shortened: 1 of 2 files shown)


benbryant0:
Closing in favour of #936 as that PR fixes the same issues but they also caught a few things that I missed.

PR #932, opened 2023-07-10
Default installation as a dev dependency

luciomartinez:
The reason behind it is to make the default what's likely the most used way to install the library. It'll help reducing the overhead for library creators of [manually doing it afterwards](https://styled-components.com/docs/faqs#marking-styled-components-as-external-in-your-package-dependencies).

### DIFF
--- a/sections/basics/installation.mdx
+++ b/sections/basics/installation.mdx
@@ -4,10 +4,10 @@ Installing styled-components only takes a single command and you're ready to rol
 
 ```
 # with npm
-npm install styled-components
+npm install --save-dev styled-components
 
 # with yarn
-yarn add styled-components
+yarn add --dev styled-components
 ```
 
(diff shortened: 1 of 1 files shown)


luciomartinez:
@probablyup what do you think?

probablyup:
I don't think so, it should go in `dependencies` if it's used in production.

luciomartinez:
> I don't think so, it should go in `dependencies` if it's used in production.

You're right in the sense that whatever is used in production should be part of `dependencies`. 
But in the case of styled-components, it's a library used at build time to generate the assets (HTML/CSS/JS) that will then go into production. 
Hence, the library itself won't be used in production. Unless I'm missing something 🤔

probablyup:
> > I don't think so, it should go in `dependencies` if it's used in production.
> 
> You're right in the sense that whatever is used in production should be part of `dependencies`. But in the case of styled-components, it's a library used at build time to generate the assets (HTML/CSS/JS) that will then go into production. Hence, the library itself won't be used in production. Unless I'm missing something 🤔

You are, s-c has a runtime that is included in the bundle if React rehydration takes place (99.9% of the time this is the case.)

PR #935, opened 2023-08-09
Update passed-props.mdx

mrdulin:
1. Fixed the prop name, `inputColor` should be `$inputColor`
2. Fixed grammer.

### DIFF
--- a/sections/basics/passed-props.mdx
+++ b/sections/basics/passed-props.mdx
@@ -25,4 +25,4 @@ render(
 );
 ```
 
-Note how the `$inputColor` prop is not passed to the DOM, but `type` and `defaultValue` are. That is styled-components are smart enough to filter non-standard attributes automatically for you.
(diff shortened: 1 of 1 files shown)


sections/basics/passed-props.mdx
```diff
@@ -25,4 +25,4 @@ render(
 );
 ```
 
-Note how the `inputColor` prop is not passed to the DOM, but `type` and `defaultValue` are. That is styled-components being smart enough to filter non-standard attributes automatically for you.
+Note how the `$inputColor` prop is not passed to the DOM, but `type` and `defaultValue` are. That is styled-components are smart enough to filter non-standard attributes automatically for you.
```
probablyup, line 28: Improved sentence

```suggestion
Note how the `$inputColor` prop is not passed to the DOM, but `type` and `defaultValue` are? The `styled` function is smart enough to filter non-standard attributes automatically for you.
```

probablyup (approved):

PR #936, opened 2023-08-09
Update attaching-additional-props.mdx

mrdulin:
1. Fixed prop name error, `size` should be `$size`
2. Fixed grammar
3. Fixed TS generic type error.


probablyup (dismissed):
Thank you :)

probablyup (approved):

PR #939, opened 2023-08-22
Update v6 migration guide for shouldForwardProp

bcole808:
In v5, prop forwarding behaved differently depending on if the target was an HTML element, or a React Component.  

This PR updates the code example showing how to re-implement the default behavior that existed in v5 in a way that more closely matches the v5 behavior.

### DIFF
--- a/sections/faqs/migration-v6.mdx
+++ b/sections/faqs/migration-v6.mdx
@@ -34,10 +34,20 @@ import { StyleSheetManager } from 'styled-components';
 
 function MyApp() {
     return (
-        <StyleSheetManager shouldForwardProp={isPropValid}>
+        <StyleSheetManager shouldForwardProp={shouldForwardProp}>
             {/* other providers or your application's JSX */}
         </StyleSheetManager>
     )
+}
+
+// This implements the default behavior from styled-components v5
+function shouldForwardProp(propName, target) {
+  if (typeof target === "string") {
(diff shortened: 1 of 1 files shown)


woodreamz:
You're right, this section is misleading, your proposal is the real v5 behavior according to the [v5 source code](https://github.com/styled-components/styled-components/blob/legacy-v5/packages/styled-components/src/models/StyledComponent.js#L140):

```javascript
const isTargetTag = isTag(elementToBeCreated); // ======> typeof elementToBeCreated === 'string'
  const computedProps = attrs !== props ? { ...props, ...attrs } : props;
  const propsForElement = {};

  // eslint-disable-next-line guard-for-in
  for (const key in computedProps) {
    if (key[0] === '$' || key === 'as') continue;
    else if (key === 'forwardedAs') {
      propsForElement.as = computedProps[key];
    } else if (
      shouldForwardProp
        ? shouldForwardProp(key, validAttr, elementToBeCreated)
        : isTargetTag
        ? validAttr(key) // ======> validAttr = isPropValid 
        : true
    ) {
      // Don't pass through non HTML tags through to HTML elements
      propsForElement[key] = computedProps[key];
    }
  }
```

woodreamz:
Also noticed `shouldForwardProp` is not documented on [StyleSheetManager section](https://styled-components.com/docs/api#stylesheetmanager).

probablyup (approved):
Thanks, agreed :)

PR #951, opened 2023-10-03
fixed visual syntax error for basics#attaching-additional-props

bigkho:
Fixed visual syntax error that failed to show the correct inputs for the docs/basics#attaching-additional-props page.

This is what the page looks live:
error

There is now no syntax error, and the desired input fields are displayed on the page:
image


probablyup (approved):
Ah thank you :)

PR #959, opened 2023-12-27
Add description of react-select custom components in migration guide v6

takurinton:
Added to the v6 migration guide for content at https://github.com/styled-components/styled-components/issues/4074#issuecomment-1870027194.


takurinton:
Perhaps the current interface is not the final look we want and the documentation and migration guide will need to be updated as changes are made.

Dron007:
I don't think it is clear what should be changed when themes are not used at all. The example in sandbox doesn't explain it, it is too cumbersome and covers specific case.

PR #960, opened 2024-01-16
docs: update the css-prop installation guide for users using v6

peter-byun:
This PR updates the `css-prop` API reference's [Usage with TypeScript section](https://styled-components.com/docs/api#usage-with-typescript) to inform readers using higher versions than `styled components` v6 that they can simply import the `CssProp` type from the `styled-component` package rather than installing a separate types package from [Definitely Typed](https://github.com/DefinitelyTyped/DefinitelyTyped).

Here is the updated version of the CSS Prop section: https://styled-components-ns1cyl5l8-styledcomponents.vercel.app/docs/api#usage-with-typescript


quantizor (approved):
Thanks!

PR #968, opened 2024-04-23
Add link for an explanation about "resolutions"

hirokifj:
Hi, I added a link for an explanation about "resolutions".
This update would make it easier to understand what "resolutions" is, as it is not obvious to those who are not familiar with yarn or npm.


quantizor (approved):
Good idea

